Issue #10, opened 2020-01-17
Training job_logs not saved correctly


garciagenrique:
Both training and testing jobs logs are named in the same way (i.e., job#.e and job#.o). 

Training jobs are sent before and there is no counter that takes into account these `n_num` jobs sent, so that when the testing jobs are sent, they crush the training logs. 

Will take care of updating the scripts.

garciagenrique:
Solved with the implementation of #11

PR #11, opened 2020-02-03
Workflow r0 dl3 - stage 1

garciagenrique:
Modification of the 4 main scripts of the full workflow so that they can be run independently AND within the script that runs the full workflow (to be pulled when this stage is revised).

Documentations has been added. Typos, PEP8 cleaning and misleading variables have also been  modified.

Minor points are identified and marked as #TODO (job management, few functions not working from `data_management.py`). Will be solved in next iterations. 

Issue #10 is solved in this PR.


garciagenrique:
(Stage 2).
Script to run the full workflow added to the PR. 

Call the four previous scripts and manage the full workflow, keeping logs and passing through the different stages the jobids use as dependencies.

lst_scripts/onsite_mc_dl1_to_dl2.py
```diff
@@ -98,7 +98,7 @@ def main(input_dir, path_models=None, config_file=None, flag_full_workflow=False
                 batch_cmd += ' --dependency=afterok:' + wait_ids_proton_and_gammas
             batch_cmd += ' --wrap="{}"'.format(cmd)
 
-            jobid_dl1_to_dl2 = os.popen(batch_cmd).read().split('\n')
```
vuillaut: rather than reading the `job_id` from the display, it is set in the variable `SLURM_JOB_ID` by the job manager.

lst_scripts/onsite_mc_r0_to_dl1.py
```diff
@@ -213,13 +216,16 @@ def main(input_dir, config_file=None, train_test_ratio=0.25, random_seed=42, n_f
...
+                jobid = os.popen(cmd).read().strip('\n')
+                jobids_r0_dl1.append(jobid)
 
                 # Fill the dictionaries if IN workflow mode
-                jobid2cmd[jobid] = cmd
-                jobid2outfile[jobid] = jobo
-                jobid2errfile[jobid] = jobe
-                jobid2partype[jobid] = DL0_DATA_DIR.split('/')[-2]  # Hardcoded, maybe if with 4 elif s ?
+                jobid2log[jobid] = {}
+                jobid2log[jobid]['particle'] = DL0_DATA_DIR.split('/')[-2]  # Hardcoded, maybe if with 4 elif s ?
```
vuillaut: you can also look for the words `gamma`, `gamma_diffuse`, `proton`, `electron` in the path...

lst_scripts/onsite_mc_dl1_to_dl2.py
garciagenrique: but then the variable must be recovered somehow (`echo $SLURM_JOB_ID`) and pass it back to the main code...
I don't see the advantage of this method (to be discussed of course !)

vuillaut: The advantage would be to not rely on display (avoiding to catch another message for example)

lst_scripts/onsite_mc_r0_to_dl1.py
vuillaut: Note: careful with `gamma` and `gamma_diffuse`...

lst_scripts/batch_dl1_utils-merge_and_copy.py
```diff
@@ -0,0 +1,44 @@
+#!/usr//bin/env python3
+
+# E. Garcia, 12/02/2020
+
+import argparse
+from .data_management import move_dir_content, check_and_make_dir_without_verification
```
vuillaut: `from .data_management` will not work.
Plus, with installed packaged, you might prefer `from lst_scripts.data_management import ... `

lst_scripts/onsite_mc_r0_to_dl1.py
```diff
@@ -115,66 +162,123 @@ def main():
...
 
+    else:
+        print("\n ==== END {} ==== \n".format(sys.argv[0]))
 
-if __name__ == '__main__':
-    main()
 
+if __name__ == '__main__':
+    args = parser.parse_args()
+    main(args.input_dir,
```
vuillaut: how can user  set `flag_full_workflow` ?

lst_scripts/onsite_mc_r0_to_dl1.py
```diff
@@ -115,66 +162,123 @@ def main():
...
-            cmd = 'sbatch -e {} -o {} {} {}'.format(jobe, jobo, base_cmd, os.path.join(dir_lists, file))
 
-            # os.system(cmd)
-            print(cmd)
+            # recover or not the jobid depending of the workflow mode
+            if not flag_full_workflow:
+                cmd = 'sbatch -e {} -o {} {} {}'.format(jobe, jobo, base_cmd, os.path.join(dir_lists, file))
+
+                print(cmd)
+                # os.system(cmd)
```
vuillaut: default ?

lst_scripts/batch_dl1_utils-merge_and_copy.py
garciagenrique: Right for the `.data_management import ...`. It was set like this because pycharm (my local program) doesn't like find the path otherwise. So they must be changed in every file.

lst_scripts/onsite_mc_r0_to_dl1.py
garciagenrique: `flag_full_workflow` is supposed to be set modified ONLY at the `r0_to_dl3` script. Why ? Because by default is set to `False` - what allows each single stage to be run separately -, and this way avoids the user to take into account all the rest of `fla_full_workflow = True` dependent arguments - what simplifies the run of each stage separately.

lst_scripts/onsite_mc_r0_to_dl1.py
garciagenrique: yes, I think as it was originally. (maybe to avoid launching 200 jobs by error when testing the code ?)

lst_scripts/onsite_mc_r0_to_dl1.py
vuillaut: got it thanks

PR #17, opened 2020-03-26
make codes executable and add source_env to dl1_dl2

garciagenrique:
- make `reorganize_dl1_output_files_hiperta_hipecta.py` and `hiperta_r0_to_dl1lstchain.py` execuatable (change the 'main():' and how the arguments are passed)
- add and update the `source_env` on `train_pipe` and `dl1_dl2` to avoid that different environments could run consecutive jobs


garciagenrique:
Just given to the re-organizer script a consistent name `reorganize_dl1hiperta_to_dl1lstchain.py` (and change where it is invoked)

PR #21, opened 2020-04-06
fix stacking of events and images

vuillaut:


garciagenrique:
Thanks for the implementation @vuillaut !

PR #22, opened 2020-04-06
Fix telid

vuillaut:


vuillaut:
My bad I forgot to finish the PR earlier

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
```diff
@@ -175,10 +171,6 @@ def modify_params_table(table, position_iterator):
     with np.errstate(invalid='ignore'):
         table.add_column(table['width'] / table['length'], name='wl')
 
-    if position_iterator == 0:
-        print("\n\tRuntime Warnings have been ignored to avoid repeated stdout prints.")
-        print("\tRuntimeWarnings due to invalid values and divide by zero in `log10(intensity)`")
-        print("\t operations and divide by zero in `wl` divisions.")
```
garciagenrique, line 181: I agree that this not super useful, but at least it does reminds the user what's going on... - if ever we are not us the users

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
```diff
@@ -197,11 +189,29 @@ def stack_by_telid(dl1_pointer):
...
     """
 
-    tabs = [Table(tel.parameters.read()) for tel in dl1_pointer]
-    stacked_param = vstack(tabs)
+    tels_params = [Table(tel.parameters.read()) for tel in dl1_pointer]
+    try:
+        tel_ids = [tel['telId'][0] for tel in dl1_pointer]
+    except:
+        # if the tel_id column does not exist, we assign tel ids by simple iteration
+        tel_ids = [i+1 for i in range(len(tels_params))]
```
garciagenrique, line 197: Is this case for hipecta files ? Not other way of retrieving the `tel_id` inside the file ?

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
vuillaut: Well then just don't use `errstate` - the warning should appear only once.

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
vuillaut: For hipecta files I don't know 🤷‍♂️
Any idea ?

vuillaut:
> Thanks for the corrections @vuillaut !
> 
> Just conform please that
> `table_dl1 = join(table_dl1, mc_event, keys='event_id')`
> will join correctly with the changes done and that there is no need of adding `tel_id` as a key for joining too.

`mc_event` does not have `tel_id` as key, right?

`['event_id', 'tel_id']` should be used to join for example the parameters table and the images table

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
garciagenrique: Before, due to the `for` for each parameter_table, the message appeared 4 times. 
Indeed, if you can take out the `np.errstate`, the warning should will either in the stdout or in the `*.o` file

lst_scripts/reorganize_dl1hiperta_to_dl1lstchain.py
garciagenrique: Without a current and bug-corrected version of hipecta, I cannot make any suggestions. 
However, if it is not for the hipecta file, why the `try, except` ? The parameter's table must contain always a `Tel_*['telId]` argument right ? or is it just for a general case and to be sure that next stages won't crash ?

vuillaut: Yes it is for hipecta files but is a generic solution.

PR #39, opened 2020-06-16
Update reorganizer and test for it 

garciagenrique:
Still need to be debugged (`default` infile need to be changed too).


garciagenrique:
Moving test to HiPeCTA repo

PR #66, opened 2020-12-18
Create lstchain_mc_dl2_to_irfs.py

vuillaut:


garciagenrique:
🥳 🥳 🥳 🥳 
Speaking about the tree path, a do you think that saving the irfs into a `IRFs` directory (at the same level that `running_analysis`, `dl1` etc...) will be ok ? inside `analysis_logs` maybe ? Or would you prefer to keep them "home made" for the moment and save them elsewhere ?

garciagenrique:
Will 🧐🧐🧐  the code now carefully 😛

garciagenrique:
@vuillaut can we then close this PR  ?

vuillaut:
Replaced by PR #89

PR #85, opened 2021-04-16
IRFs output filename

garciagenrique:
Add gamma type - point like or diffuse, and the gamma_offset to the output filename. Otherwise, this info is only recovered either from the config or the cmd sent to slurm.


garciagenrique:
@vuillaut or should all the IRFs  be launched in this stage ? 
By all I mean 
 - gamma point like 0.0deg
 - gamma point like 0.4deg
 - gamma diffuse
  ?

PR #88, opened 2021-04-21
plot models features importance

vuillaut:
to add after the model training


garciagenrique:
Not an experte with the backend but not specifying none we are at 
```
real    0m39.443s
user    0m17.072s
sys     0m24.077s
```
while if we use the `'matplotlib.use('Agg')` backend we down to
```
real    0m23.104s
user    0m16.228s
sys     0m20.287s
```

any suggestions @vuillaut ?

maxnoe:
Don't specify backends in code, use the environment variable `MPLBACKEND` or the `matplotlibrc` file.

On clusters or in any situation without interactivity, always use `agg`.

Setting the backend from within the code is error-prone (since it has to be done before anything imports `matplotlib.pyplot`) and breaks interactivity in case someone actually wants to look at a plot.

garciagenrique:
> On clusters or in any situation without interactivity, always use `agg`.
> 

Thanks @maxnoe ! Precisely this stage is intended to be batched into the cluster too. 
I will then use a `export MPLBACKEND=Agg` before the command to be batched.

garciagenrique:
Hi @LukasNickel, @maxnoe,

As new 'official' contributors to the repository (😛 ), let me know if you would like to review the PR. Feel free to say no of couse ;)

vuillaut:
I can't approve because I authored part of the PR but your changes look good to me.

PR #93, opened 2021-04-22
adding export env function for provenance

vuillaut:


lstmcpipe/config/__init__.py
```diff
@@ -0,0 +1,10 @@
+import os
+from shutil import which
+
+
+def export_env(outdir='.'):
```
maxnoe, line 5: I also investigated yesterday how to best do this. There is `pkg_resources.working_space` which includes all packages with name, version, path that are accessible to the current interpreter.:

```
from pkg_resources.working_set


for package in sorted(working_set, key=lambda p: p.project_name):
     print(package.project_name, package.version, package.module_path)
```

vuillaut: awesome, thank you!

garciagenrique (approved):

garciagenrique:
For the record,
 - pytables
 - numpy
 - astropy
 - pandas
 - hiperta

are needed in the package (but only for the hiperta 'module', although this will disappear when we get rid of the lst-rta converter)

(pytables, numpy, astropy and pandas should come with lstchain)

PR #99, opened 2021-04-29
update version at __init__ file

garciagenrique:


lstmcpipe/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.4"
+__version__ = "0.41"
```
maxnoe, line 1: Missing dot?

garciagenrique: Thanks max. I was too quick....

PR #100, opened 2021-05-01
fix path to images in DL1 files

vuillaut:
see:
https://gitlab.in2p3.fr/CTA-LAPP/rta/HiPeRTA_Stream/-/merge_requests/5/diffs

### DIFF
--- a/lstmcpipe/hiperta/reorganize_dl1hiperta300_to_dl1lstchain060.py
+++ b/lstmcpipe/hiperta/reorganize_dl1hiperta300_to_dl1lstchain060.py
@@ -51,7 +51,7 @@ def stack_and_write_images_table(input_filename, hfile_out, node_dl1_event):
     dump_plus_copy_node_to_create_new_table(input_filename,
                                             hfile_out,
                                             image_table,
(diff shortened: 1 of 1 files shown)


garciagenrique (commented):
Thanks @vuillaut. 

Wouldn't it have more sense to update the reorganiser of `lstmcpipe` with the same as `hiperta_stream` ? (yes, double work, but we will avoid this kind of problems in the future)

vuillaut:
> Thanks @vuillaut.
> 
> Wouldn't it have more sense to update the reorganiser of `lstmcpipe` with the same as `hiperta_stream` ? (yes, double work, but we will avoid this kind of problems in the future)

I don't understand, isn't it precisely what I just did (see pasted link)?

garciagenrique:
Sorry, you are completely right. 

https://gitlab.in2p3.fr/CTA-LAPP/rta/HiPeRTA_Stream/-/blob/master/hiperta_stream/reorganizer_dl1hiperta300_stream_to_dl1lstchain063.py#L113

Don't know why it was changed. In any case it should be `image`. Thanks for catching this.

garciagenrique (approved):

PR #120, opened 2021-07-07
Fix relative import to work after installation

LukasNickel:


garciagenrique:
Thanks for the catch @LukasNickel !

garciagenrique (approved):

PR #121, opened 2021-07-15
Refactor stages script to take out the 'scripting' functionality

garciagenrique:
- Takes out from all the onsite_* scripts the possibility to be run interactively (nobody was using this functionality. Also it was misleading 'cause all the scripts were copied in the outdir. Users though that this was the script to be used to reprocess data. Indeed it is, but orchestrated in  the lstmpipe framework, so not immediate to be used.)
- Creates a `stages` submodule were to store all the stages --> #TODO make everything much more modulable !
- r0_to_dl1 stage won't be refactored until #111 is merged.


lstmcpipe/workflow_management.py
```diff
@@ -9,12 +9,13 @@
 import pprint
 import calendar
 import lstchain
+from lstmcpipe.stages import (onsite_mc_merge_and_copy_dl1,
+                              onsite_mc_train,
+                              onsite_mc_dl2_to_irfs,
+                              onsite_mc_dl1_to_dl2
+                              )
```
LukasNickel, line 16: If you import the functions from the subscripts into `stages/__init__.py` and make them available via `__all__`, you could do `from lstmcpipe.stages import r0_to_dl1, dl1_to_dl2, ...` and directly use them avoiding the lengthy calls in the batch functions.

E.g. like here
https://github.com/cta-observatory/ctapipe/blob/master/ctapipe/core/__init__.py

garciagenrique: Thanks for the tips and the review ! Will commit it soon !

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
```diff
@@ -0,0 +1,278 @@
...
+
+    # 2. check that all files have been created in DL1 based on training and testing lists
+    # just check number of files first:
+    if not len(os.listdir(DL1_training_dir)) == len(read_lines_file(training_filelist)):
+        tf = check_files_in_dir_from_file(DL1_training_dir, training_filelist)
+        if tf != []:
+            # TODO to fix this
+            print(f"\n\n\n ****!!!!! {len(tf)} files from the training list are not in the `DL1/training directory:\n{tf} "
+                  "Continuing\n\n\n")
+        # if tf != [] and not flag_full_workflow:
```
LukasNickel, line 149: Not relevant anymore

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
```diff
@@ -0,0 +1,278 @@
...
+        #     query_continue("{} files from the training list are not in the `DL1/training` directory:\n{} "
+        #                    "Continue ?".format(len(tf), tf))
+
+    if not len(os.listdir(DL1_testing_dir)) == len(read_lines_file(testing_filelist)):
+        tf = check_files_in_dir_from_file(DL1_testing_dir, testing_filelist)
+        if tf != []:
+            # TODO to fix this
+            print(f"\n\n\n ****!!!!! {len(tf)} files from the testing list are not in the `DL1/testing directory:\n{tf} "
+                  "Continuing\n\n\n")
+        # if tf != [] and not flag_full_workflow:
```
LukasNickel, line 159: Same as above
`flag_full_workflow` was removed

lstmcpipe/stages/onsite_mc_train.py
```diff
@@ -0,0 +1,92 @@
...
+        jobid of the batched job to be send (for dependencies purposes) to the next stage of the
+        workflow (onsite_mc_dl1_to_dl2)
+
+    models_dir : str
+        Path with the directory where the models are stored
+
+    """
+    log_train = {}
+
+    # dl1_gamma_dir = os.path.dirname(os.path.abspath(gamma_dl1_train_file))
```
LukasNickel, line 49: Not used

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
```diff
@@ -143,22 +143,14 @@ def merge_dl1(input_dir, particle2jobs_dict, particle=None, flag_merge=False, fl
...
         tf = check_files_in_dir_from_file(DL1_training_dir, training_filelist)
         if tf != []:
-            # TODO to fix this
-            print(f"\n\n\n ****!!!!! {len(tf)} files from the training list are not in the `DL1/training directory:\n{tf} "
-                  "Continuing\n\n\n")
-        # if tf != [] and not flag_full_workflow:
-        #     query_continue("{} files from the training list are not in the `DL1/training` directory:\n{} "
-        #                    "Continue ?".format(len(tf), tf))
+            raise Exception(f" * {len(tf)} files from the training list are not in the `DL1/training directory:\n{tf}"
+                  "\nContinuing.")
```
LukasNickel, line 147: You are not really continuing if you raise an Exception, right?

LukasNickel: It was just printing the warning before, so this is a change in behaviour.
is that intended?

garciagenrique: No intended indeed...^^ thanks for the catch
How would you proceed (also as an user) @LukasNickel ? If the tests (checking that all the files that were processed are indeed gonna be merged) doesn't pass, would you stop the pipeline or would you just raise a warning (and in the clos future store it in the log ?)

LukasNickel: As a user I would probably suspect, that the pipeline stops at that point. I could then debug the issue and once it is resolved, continue with my analysis.
If the output looks alright, nobody is going to look into the logs for a while, so nobody might notice that there is an error in the files.

If its just one broken file, it would be nice if the chain could continue where it left off, once that has been fixed. But that might be a bit tricky (and maybe the separation into datalevels is enough for that)

lstmcpipe/stages/mc_dl2_to_irfs.py
```diff
@@ -154,7 +110,7 @@ def main(dl2_directory, config_file, irf_point_like=True, irf_gamma_offset='off0
     log_dl2_to_irfs = {}
     list_job_id_dl2_irfs = []
 
-    if not flag_full_workflow or log_from_dl1_dl2 == {}:
+    if log_from_dl1_dl2 == {}:  # Empty dict and thus no dl2 path files
```
LukasNickel, line 113: Since `log_from_dl1_dl2` does not have a default value anymore, is this this a reasonable check or should it be 
```python
if not log_from_dl1_dl2:
    ....
```
?

lstmcpipe/stages/mc_r0_to_dl1.py
```diff
@@ -320,75 +244,50 @@ def main(input_dir, config_file=None, train_test_ratio=0.5, random_seed=42, n_r0
...
+
+            # Fill the dictionaries if IN workflow mode
+            jobid2log[jobid] = {}
+            jobid2log[jobid]['particle'] = particle
+            jobid2log[jobid]['set_type'] = set_type
+            jobid2log[jobid]['jobe_path'] = jobe
+            jobid2log[jobid]['jobo_path'] = jobo
+            jobid2log[jobid]['sbatch_command'] = cmd
+
+            # print(f'\t\t{cmd}')
```
LukasNickel, line 274: We should really just move to logging and mark those as debug :D
But that can be done in another PR

setup.py
```diff
@@ -31,6 +31,8 @@ def readfile(filename):
         'lstmcpipe_plot_models_importance = lstmcpipe.plots.plot_models_importance:main',
         'lstmcpipe_plot_irfs = lstmcpipe.plots.plot_irfs:main',
         'lstmcpipe_dl2_to_sensitivity = lstmcpipe.stages.script_dl2_to_sensitivity:main'
+        'lstmcpipe_utils_move_dir = lstmcpipe.stages.script_merge_utils_move_dir:main',
```
LukasNickel, line 34: Should these have their own entry points?
Looks like they could just be functions, that are called from the stages or main script

lstmcpipe/stages/mc_dl1_to_dl2.py
```diff
@@ -0,0 +1,124 @@
...
+        wait_jobs = ''
+    else:
+        wait_jobs = ','.join([wait_jobid_train_pipe, wait_jobids_merge])
+
+    job_name = {'electron': 'dl1-2_e',
+                'gamma': 'dl1-2_g',
+                'gamma-diffuse': 'dl1-2_gd',
+                'proton': 'dl1-2_p',
+                'gamma_off0.0deg': 'g.0_1-2dl',
+                'gamma_off0.4deg': 'g.4_1-2dl'
```
LukasNickel, line 80: Maybe these could be changed to be more consistent, e.g. particle name in front for all cases

lstmcpipe/stages/mc_r0_to_dl1.py
garciagenrique: Better another PR, yes ;)

lstmcpipe/stages/mc_dl2_to_irfs.py
garciagenrique: I think both are correct. Checking that a dict is empty or that the var does not exists. Won't be able to justify more one or another ^^

setup.py
garciagenrique: I think they should.
These stages are quite useless (I'm sure this can be done somehow), but I didn't find another way of performing these actions (move DL1 files, copy config and move DL1 dir) **after** the merging has succeeded. If they are performed before, the pipeline will not work because no "final" DL1 file has been merged. Also, the end of these jobs will allow the training to start (If no 'final' DL1 file in the correct directory, the training won't work).

garciagenrique: The only solution to solve this (I think) is to completely remove the `running_analysis` stage (comes from some years ago), and create and merge everything in the same DL1 dir.

Maybe another point to improve ? ^^

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
garciagenrique: > As a user I would probably suspect, that the pipeline stops at that point. I could then debug the issue and once it is resolved, continue with my analysis.
The pipeline can only stops when it gets to this point (stages will take time to finish). But this messages will appear before because the filelist are written when the pipeline is launched although the files are not processed yet. 


> If its just one broken file, it would be nice if the chain could continue where it left off, once that has been fixed. But that might be a bit tricky (and maybe the separation into datalevels is enough for that)

This is quite difficult to do indeed... (I think 😄 ). I don't know how to put some jobs in "standby" inside a cluster if they fail  and re-launch them later. At least we will need to think how to do it very carefully.

lstmcpipe/stages/mc_dl2_to_irfs.py
LukasNickel: Yeah, but if you pass an empty list or just `None` to the function now, it wont trigger

lstmcpipe/stages/mc_dl1_to_dl2.py
LukasNickel: What I mean is that for offset gammas the particle comes first and for other particles it comes after the dl1-2 part

setup.py
LukasNickel: Ok, then lets keep it like that for now

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
LukasNickel: Wait, the message appears before?
It should process files, then make sure all files are processed and then print a warning or fail, right?
Of course it cant fail before it processes the dl1 stage, why woul dthe message be printed before?

setup.py
LukasNickel: There is a comma missing in l33 btw.
The script is not getting installed

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
garciagenrique: Hmm.. i'm indeed not sure what this part is really doing (it should not work as the sub-dl1 files are not processed and created yet, so this check should always fail...)

Another point in the todo list ;-) improve the merge stage 😬

LukasNickel (approved):

lstmcpipe/stages/onsite_mc_merge_and_copy_dl1.py
garciagenrique: OK, I have to recognise that i'm not sure how this check was not triggering before.... it will always trigger indeed. I think that I would just take it out 'cause it will never be satisfied, indeed.

garciagenrique: And do some tests with master version in case I'm missing/messing something big

garciagenrique: Again, confused for all the changes (and doing 3 things at the same time...^^).

The original code was 
```
   if tf != [] and not flag_full_workflow:
```

so of course this check was not happening. Now that there is no `flag_full_workflow` so this check will always 'fail'. Taking out the check then.

Thanks for the catch @LukasNickel

garciagenrique:
Sorry @LukasNickel, just found couple of minor changes. You have already review them ;)

PR #123, opened 2021-07-22
Refactor bash files

garciagenrique:
Substitute core bash files by python scripts. 
Renaming and removing files, and create corresponding entry points.


lstmcpipe/scripts/script_merge_utils_copy_config.py
```diff
@@ -0,0 +1,33 @@
...
+                    help='destination argument of move_dir_content',
+                    )
+
+
+def main():
+    args = parser.parse_args()
+
+    # HiPeRTA config is *.txt (<v4.0.0), otherwise is *.yml
+    config_files = [os.path.join(args.source, f) for f in os.listdir(args.source)
+                    if f.endswith(('.json', '.txt')) or f.startswith('hiperta_conf*.yml')]
```
LukasNickel, line 26: Does this expression make sense?
This selects json, txt and yml (only if it has hiperta in its name).
It specifically avoids other yml configs, why is that? The main config is yml as well.

garciagenrique: Hiperta config can be `.txt` or `.yml` (hiperta still supports both configs), and lstchain/ctapipe `.json`.

This expression looks for certain config files (the ones used in the r0_dl1 + merge stage and to be moved to `DL1` dir), to "maintain" a kind of provenance and copy them in the same dir as the merged DL1 files. (The merge stage is indeed an extension of r0_to_dl1, so basically to indicate the config used for this stage).

I think that copying the main (pipeline) configuration file in the DL1 dir is not really interesting, that's why it doesn't search for all the *.yml files.

garciagenrique:
PR is ready. 
I'm working on testing the 'core' python scripts, but might open a new PR.

lstmcpipe/scripts/script_merge_utils_copy_config.py
LukasNickel: I think in the future it would be best to explicitly mention the files and cal the script in each stage instead of these conditions. Maybe some configs will be skipped if any of the projects decides to use another format.
But I guess its fine for now

lstmcpipe/stages/mc_merge_and_copy_dl1.py
```diff
@@ -14,8 +14,8 @@
 import os
 from lstmcpipe.io.data_management import (
     check_job_logs,
-    read_lines_file,
-    check_files_in_dir_from_file
+    # read_lines_file,
```
LukasNickel, line 17: Could be removed

LukasNickel (approved):

LukasNickel (changes requested):
Actually, I havent been able to get this to run.
Have you tested everything?

LukasNickel:
There was an issue with the subprocess call, that should now be fixed.
Also the path to the r0 list files was not complete (Why did you change that?)

garciagenrique:
> Also the path to the r0 list files was not complete (Why did you change that?)

Probably mistake from my side 😅 . I will be doing the tests in a moment !

LukasNickel (approved):

PR #124, opened 2021-07-22
Cleanup config code

LukasNickel:
Prerequisites: #123

Simplifies some code and adds unit tests.
There is one small change in behaviour:
If you want to avoid setting a `prod_id` in the config, you can just omit the key now. I think this is nicer than specifying the key and setting no value.

Everything else stays the same, diff looks huge until #123 is merged as those commits show up as well.


lstmcpipe/config/pipeline_config.py
```diff
@@ -0,0 +1,229 @@
...
+    src_env = (f"source {loaded_config['source_environment']['source_file']}; "
+               f"conda activate {loaded_config['source_environment']['conda_env']}; "
+               )
+    config['source_environment'] = src_env
+
+    # 3 - particles loop
+    config['source_environment'] = src_env
+    config['all_particles'] = particles
+
+    # 3.1 - Gammas' offsets
```
garciagenrique, line 180: It can be removed or moved to line 195

PR #125, opened 2021-08-31
Logging + X

LukasNickel:
Merge after #123 and #124

- print ->  logging everywhere
- Same string formatting in all logging calls (.format() instead of f-strings)

- Job arrays for the r0->dl1 stage. This avoids spawning 1000s of jobs and makes job allocation easier. The maximum number of parallel runs is set to 20 (per particle type), which might be pretty low? We could also make this configurable if we want to
- batch_stage functions got moved to the stages in order to simplify the `workflow_management` module


.gitignore
```diff
@@ -9,4 +9,6 @@ build
 .idea
 __pycache__/
 lstmcpipe/__pycache__/
-.coverage
\ No newline at end of file
+.coverage
+# vim swap files
+*.swp
```
garciagenrique, line 14: Add `.pylintrc` to the .gitignore

lstmcpipe/stages/mc_r0_to_dl1.py
```diff
@@ -202,6 +288,7 @@ def r0_to_dl1(input_dir, config_file=None, train_test_ratio=0.5, random_seed=42,
     jobids_r0_dl1 = []
 
     for set_type in 'training', 'testing':
+        log.debug("Generating list for {} step")
```
garciagenrique, line 291: I think you forgot a .format with the `set_type` of the loop

lstmcpipe/stages/mc_r0_to_dl1.py
```diff
@@ -213,69 +300,62 @@ def r0_to_dl1(input_dir, config_file=None, train_test_ratio=0.5, random_seed=42,
...
-            if particle == 'proton':
-                queue = 'long'
-            else:
-                queue = 'short'
+        files = [f.resolve().as_posix() for f in Path(dir_lists).glob("*")]
 
-            cmd = f'sbatch --parsable -p {queue} -J {job_name[particle]} -e {jobe} -o {jobo} ' \
-                  f'--wrap="{base_cmd} -f {os.path.join(dir_lists, file)} -o {output_dir}"'
+        if set_type == 'training':
+            jobo = os.path.join(JOB_LOGS, "job_%A_%a_train.o")
```
garciagenrique, line 319: Are the `%A` and `%a` strings for a later formatting ? I think they are not formatted when passed to the final `cmd`

garciagenrique: I mean, the final slurm command looks like this :
```
sbatch --parsable --array=[0-0]%20 -p short 
-e {...}/20210907_v0.7.3_prod5_trans_80_test_job_arrays_logging/off0.0deg/job_logs/job_%A_%a_train.e 
-o {...}/20210907_v0.7.3_prod5_trans_80_test_job_arrays_logging/off0.0deg/job_logs/job_%A_%a_train.o 
-J g0.0_LST_r0dl1  --wrap="  ... "
```

LukasNickel: They are supposed to be filled by the slurm task with the array and task id.
See:
https://slurm.schedmd.com/job_array.html

lstmcpipe/stages/mc_r0_to_dl1.py
```diff
@@ -213,69 +300,62 @@ def r0_to_dl1(input_dir, config_file=None, train_test_ratio=0.5, random_seed=42,
...
+            jobe = os.path.join(JOB_LOGS, "job_%A_%a_test.e")
+
+        if particle == 'proton':
+            queue = 'long'
+        else:
+            queue = 'short'
 
-            jobid = os.popen(cmd).read().strip('\n')
-            jobids_r0_dl1.append(jobid)
+        slurm_options = f"--array=[0-{len(files)-1}]%{n_jobs_parallel} "
```
garciagenrique, line 330: Are the `[ ]` intended ? 
`sbatch --parsable --array=[0-0]%20`

lstmcpipe/stages/mc_r0_to_dl1.py
garciagenrique: Wow, cool ! 😄

lstmcpipe/stages/mc_r0_to_dl1.py
LukasNickel: It looks like they are not supposed to be there.
Not sure where I read it, the documentation does not mention it...
Would test it once without the brackets but it looks like they are not needed

garciagenrique (approved):

PR #126, opened 2021-09-10
Reformat everything with black

LukasNickel:
Simplified two if statements and ran black on everything.
Should not break anything, ran fine on the test set


LukasNickel:
I did not add a pre commit hook or github actions check for the future yet, but we might want to do that as well

PR #130, opened 2021-10-11
Pathlib

### LINKED ISSUE #62: Use Path library, please

LukasNickel:
Replace `os.path` with the `pathlib` equivalent
Fixes #62

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 # Licensed under a MIT license - see LICENSE.rst
 
 import os
+from pathlib import Path
 import re
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -10,61 +11,65 @@
 
 
 def find_scripts(script_dir, prefix):
-    script_list = [os.path.join(script_dir, f) for f in os.listdir(script_dir) if f.startswith(prefix)]
+    script_list = [f for f in Path(script_dir).iterdir() if f.name.startswith(prefix)]
```
garciagenrique, line 14: This comprehension list will be a list of `pathlib.Path` objects and not a list of stings. Is this intended ? I assume it does not matter a lot, but I'd prefer to check.

LukasNickel: It gets resolved in `lstmcpipe_start.py`, eg here:
https://github.com/cta-observatory/lstmcpipe/blob/d4f6c634772fc6a51649a8183f180814948b5656/lstmcpipe/lstmcpipe_start.py#L138

garciagenrique (approved):

PR #133, opened 2021-11-03
Add compare irfs script

vuillaut:


LukasNickel:
This is to compare different productions/settings, right?
Should be fine I suppose

LukasNickel (approved):

garciagenrique:
Sorry for the late review.

@vuillaut 
`main()` functions should not receive arguments and arg parser should be inside main too. 

Also, line 27
`log = logging.getLogger("lstchain MC DL2 to IRF - sensitivity curves")`
would be nice to modify the message, it's the same of the `script_dl2_to_sensitivity.py` script.

minor changes, not urgent, they will be patch in another PR.

PR #146, opened 2022-01-13
take out --smart argument from lstchain_merge_hdf5_files on lstchain …

garciagenrique:
…workflow


LukasNickel:
I would have removed it eberywhere, but if hiperta still needs it ...

LukasNickel (approved):

PR #148, opened 2022-01-18
Fix issues with sensitivity calculation and plots

vuillaut:


lstmcpipe/plots/plot_irfs.py
```diff
@@ -417,7 +417,7 @@ def main():
 
     plot_summary_from_file(args.filename)
 
-    os.makedirs(os.path.dirname(args.outfile), exist_ok=True)
+    Path(args.outfile).parent.mkdir(exist_ok=True)
```
maxnoe, line 420: You can directly give `type=Path` to the add_argument

garciagenrique (approved):

PR #149, opened 2022-01-20
pep8, add doc and allow saving fig or not in script compare irfs

garciagenrique:


lstmcpipe/scripts/script_compare_irfs.py
```diff
@@ -38,24 +55,33 @@ def main():
...
                         action='store_true',
                         dest='cta_north',
                         help='add CTA north performances curves'
                         )
 
+    parser.add_argument('--save-figure', '-s',
+                        action='store',
+                        type=lambda x: bool(strtobool(x)),
+                        dest='save_figure',
+                        help='Save resulting figure',
```
vuillaut, line 78: implement with `action='store_false'` (remove type)

vuillaut (changes requested):

vuillaut:
> Why would you not want to save the figure? These scripts are not meant for interactive use, right?
> 
> Also the way I usually see this handled is to leave the `output_file` an empty string or `False` and then do:
> 
> ```python
> if output_file:
>     fig.savefig(output_file)
> else:
>     fig.show()
> ```
> 
> Does this even show the figure in the other case? Whats the point of calling the script?



> Why would you not want to save the figure? These scripts are not meant for interactive use, right?
> 
> Also the way I usually see this handled is to leave the `output_file` an empty string or `False` and then do:
> 
> ```python
> if output_file:
>     fig.savefig(output_file)
> else:
>     fig.show()
> ```
> 
> Does this even show the figure in the other case? Whats the point of calling the script?

I agree with `output_file=None` or `False` and the proposed implementation that would be much cleaner.

Then the script args can follow the same logic (save figure only if an outfile is passed).

Concerning the interactivity, I do use the **function** (not the script obviously) in notebooks at least  🤷‍♂️

LukasNickel:
> Concerning the interactivity, I do use the **function** (not the script obviously) in notebooks at least  man_shrugging

Of course, interactive plots are nice and all.
But this is meant to be used on the cluster in a scheduled job. There is no reason for interactivity there.
(Thats my first impression, I might miss something, which is why I asked for the reason)

vuillaut:
default None for outfile in function plot_comparison

garciagenrique:
> Concerning the interactivity, I do use the function (not the script obviously) in notebooks at least man_shrugging



> > Concerning the interactivity, I do use the **function** (not the script obviously) in notebooks at least  man_shrugging
> 
> Of course, interactive plots are nice and all. But this is meant to be used on the cluster in a scheduled job. There is no reason for interactivity there. (Thats my first impression, I might miss something, which is why I asked for the reason)

Everything is indeed expected to be run at the cluster as batched job. But some examples of the school were calling the function interactively, for example :). And I have the impression that if we would like to start comparing plots, this functionality might be useful

garciagenrique:
Thanks for the changes @vuillaut !

PR #150, opened 2022-02-02
Store true

vuillaut:


garciagenrique:
Sorry the prod was launched, won't be merging, at least for the moment

LukasNickel:
Why can we not merge when a production is running?

garciagenrique:
> Why can we not merge when a production is running?

Wanted to tag before the exact version that created the HiPeRTa MC prod.

garciagenrique:
And also because the PR is missing the corresponding changes to the strings that call these entry poins.

garciagenrique (changes requested):
Change also the way these entry points are called in the rest of the code, not only how the arguments are passed.

vuillaut:
> Change also the way these entry points are called in the rest of the code, not only how the arguments are passed.

Done.
`lstmcpipe/scripts/script_batch_filelist_rta.py` is not called anywhere, right?

LukasNickel:
> > Change also the way these entry points are called in the rest of the code, not only how the arguments are passed.
> 
> Done. `lstmcpipe/scripts/script_batch_filelist_rta.py` is not called anywhere, right?

It seems to be called here:
https://github.com/cta-observatory/lstmcpipe/blob/dfaf65c5516065a02295cd55b626f3a18c82decb/lstmcpipe/stages/mc_r0_to_dl1.py#L210

lstmcpipe/stages/mc_r0_to_dl1.py
```diff
@@ -129,6 +129,7 @@ def r0_to_dl1(
     workflow_kind="lstchain",
     keep_rta_file=False,
     n_jobs_parallel=20,
+    debug_mode=False,
```
LukasNickel, line 132: I see no way to set this, how do you want to use it?

garciagenrique: hhmm right... 🤔

lstmcpipe/scripts/script_batch_filelist_rta.py
```diff
@@ -66,11 +63,13 @@ def main():
...
-                f"--debug_mode={args.debug_mode}",
+                f"--config={args.config_file}",
             ]
-            if args.config_file:
-                cmd.append(f"--config={args.config_file}")
+
+            if args.keep_file:
+                cmd.append(f"--keep_file")
+            if args.debug_mode:
+                cmd.append(f"--debug_mode")
```
LukasNickel, line 72: No f string

lstmcpipe/scripts/script_batch_filelist_rta.py
```diff
@@ -66,11 +63,13 @@ def main():
...
                 f"--outdir={args.output_dir}",
-                f"--keep_file={args.keep_file}",
-                f"--debug_mode={args.debug_mode}",
+                f"--config={args.config_file}",
             ]
-            if args.config_file:
-                cmd.append(f"--config={args.config_file}")
+
+            if args.keep_file:
+                cmd.append(f"--keep_file")
```
LukasNickel, line 70: no f string

lstmcpipe/stages/mc_r0_to_dl1.py
garciagenrique: Indeed, there are functionalities that were implemented long time ago, but now I'm not sure they make much sense.... (before you could configured HiPeRTA from the config file...). Same for `keep_files` indeed.

garciagenrique: I would propose.
Nobody uses hiperta functionality but me. 
Let's not add extra args to the config file that will never be used. 
Either I implement some optional args when `hiperta` workflow is set (and open an issue otherwise I will forgot), either (what would be most probable) I change the default value of the vars on the script by hand and I re-install 🙈 🙈 🙈 🙈  - not that I do this always 😂

garciagenrique:
Would give a though how to do things correctly, leaving the PR open for the moment.

vuillaut:
@garciagenrique ?

PR #154, opened 2022-02-08
lstmcpipe_envname in environment.yml

vuillaut:


LukasNickel:
I think thats better, right? @garciagenrique

PR #155, opened 2022-02-14
reformat dl1_process_dl1.py

garciagenrique:
Everything in a single commit... (my bad, wanted to do many things at the same time)

 - take out all capital letters 
 - substitute os.path by Path 
 - clarify dl1_files_per_job argument

### DIFF
--- a/lstmcpipe/stages/mc_process_dl1.py
+++ b/lstmcpipe/stages/mc_process_dl1.py
@@ -50,7 +50,7 @@ def batch_process_dl1(
     workflow_kind: str
         One of the supported pipelines. Defines the command to be run on r0 files
     new_production: bool
-        Whether to analysis simtels or reprocess existing dl1 files.
+        Whether to analysis simtel or reprocess existing dl1 files.
 
     Returns
     -------
@@ -177,7 +177,7 @@ def r0_to_dl1(
         If the number of r0 files found in `input_dir` is less than 100, it is consider to be a test on a small
         production. Therefore, the number of r0 files treated per batched stage (r0_to_dl1 or dl1ab) will be set to 10.
 
-        Usual productions have =>1000 r0 files, in this case, the number of files per job (r0_to_dl1 or dl1ab) will be
+        Usual productions have >=1000 r0 files, in this case, the number of files per job (r0_to_dl1 or dl1ab) will be
         fixed to 50 (in case of gamma and electrons), and 100 for protons. Because of this protons end up in the
         long queue and other particles are submitted to the short queue.
     particle : str
@@ -239,7 +239,7 @@ def r0_to_dl1(
     else:
         base_cmd = ''
         jobtype_id = ''
-        log.critical("Please, selected an allowed workflow kind.")
+        log.critical("Please, select an allowed workflow kind.")
         exit(-1)
 
     job_name = {
--- a/lstmcpipe/stages/mc_process_dl1.py
+++ b/lstmcpipe/stages/mc_process_dl1.py
@@ -8,9 +8,9 @@
 import os
 import time
 import shutil
-from numpy.random import default_rng
-from pathlib import Path
 import logging
+from pathlib import Path
+from numpy.random import default_rng
 from lstmcpipe.io.data_management import (
     check_data_path,
     get_input_filelist,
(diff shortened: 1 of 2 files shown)


lstmcpipe/stages/mc_process_dl1.py
```diff
@@ -543,17 +543,15 @@ def submit_dl1_jobs(
...
-            JOB_LOGS, f"job_%A_%a_{'train' if set_type=='training' else 'test'}.o"
-        )
-        slurm_options["error"] = os.path.join(
-            JOB_LOGS, f"job_%A_%a_{'train' if set_type=='training' else 'test'}.e"
-        )
+        slurm_options["output"] = \
+            job_logs_dir.joinpath(f"job_%A_%a_{'train' if set_type=='training' else 'test'}.o")
+        slurm_options["error"] = \
+            job_logs_dir.joinpath(f"job_%A_%a_{'train' if set_type=='training' else 'test'}.e")
         if particle == "proton":
```
LukasNickel, line 550: This can probably be removed unless we want to put them into long again.
But I dont think we need to

lstmcpipe/stages/mc_process_dl1.py
```diff
@@ -169,16 +170,16 @@ def r0_to_dl1(
...
         If the number of r0 files found in `input_dir` is less than 100, it is consider to be a test on a small
-        production. Therefore, the number of r0 files treated per batched stage will be set to 10.
+        production. Therefore, the number of r0 files treated per batched stage (r0_to_dl1 or dl1ab) will be set to 10.
 
-        Usual productions have =>1000 r0 files, in this case, the number of files per job will be fixed to 50 (in case
-        of gamma and electrons), and 100 for protons. Because of this protons end up in the long queue and other particles
-        are submitted to the short queue.
+        Usual productions have =>1000 r0 files, in this case, the number of files per job (r0_to_dl1 or dl1ab) will be
+        fixed to 50 (in case of gamma and electrons), and 100 for protons. Because of this protons end up in the
+        long queue and other particles are submitted to the short queue.
```
LukasNickel, line 182: Actually we dont differentiate anymore, see below

lstmcpipe/stages/mc_process_dl1.py
garciagenrique: Tricky point, not yet "optimised". Indeed it depends on the number of jobs that each `process_dl1` job contains. Also on the kind of workflow that we used (lstchain, hiperta, ctapipe).
Protons usually take longer, indeed I would leave them on long by default just in case.

lstmcpipe/stages/mc_process_dl1.py
garciagenrique: I think there are two different things.

It is not the same the number of batched 'r0_to_dl1' jobs, that the number of files that each of this jobs will sub-process.

n_of_parallel_jobs would determine the array length for a particle, i.e., 20, then, no more than 20 job-particle would be able to run at the same time on the cluster. However, we could have created 100 file lists, that will imply 100 particle jobs to be run.

lstmcpipe/stages/mc_process_dl1.py
garciagenrique: not always 🙈 , but indeed you're completely right.
Opening an issue to optimise this :)

lstmcpipe/stages/mc_process_dl1.py
```diff
@@ -250,33 +251,33 @@ def r0_to_dl1(
...
+    log.info("Working on DL0 files in {}".format(input_dir))
 
-    check_data_path(DL0_INPUT_DIR)
-    raw_files_list = get_input_filelist(DL0_INPUT_DIR)
+    check_data_path(input_dir)
+    raw_files_list = get_input_filelist(input_dir)
 
     if len(raw_files_list) < 100:
-        n_parallel_dl1_jobs = 10
+        dl1_files_per_job = 10
```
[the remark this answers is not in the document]
vuillaut: Moreover, this becomes obsolete with next condition line #262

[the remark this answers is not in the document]
garciagenrique: This is just for small datasets. For tests datasets indeed, in where there are 20 files, thus I divide training and testing into a single file with 10 dl1 files

PR #156, opened 2022-02-14
Slurm config account

garciagenrique:
Allow selecting the slurm account when batching a production

### DIFF
--- a/lstmcpipe/config/pipeline_config.py
+++ b/lstmcpipe/config/pipeline_config.py
@@ -207,12 +207,20 @@ def parse_config_and_handle_global_vars(loaded_config):
         f"source {loaded_config['source_environment']['source_file']}; "
         f"conda activate {loaded_config['source_environment']['conda_env']}; "
     )
-    config["source_environment"] = src_env
+    # 2.1 - Parse slurm user config account
+    if config["slurm_config"]:
+        slurm_account = config["slurm_config"]["user_account"]
+    else:
+        slurm_account = ""
+
+    # 2.2 - Create a dict for all env configuration and slurm configuration (batch arguments)
+    config["batch_config"] = {
+        "source_environment": src_env,
+        "slurm_account": slurm_account
+    }
 
     # 3 - particles loop
     config["all_particles"] = particles
-
-    # 3.1 - Gammas' offsets
 
     # 4 - Stages to be run
     config["stages_to_run"] = stages_to_be_run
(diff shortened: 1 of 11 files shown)


LukasNickel:
Tests are failing, looks good otherwise

lstmcpipe/config/pipeline_config.py
```diff
@@ -207,14 +207,21 @@ def parse_config_and_handle_global_vars(loaded_config):
         f"source {loaded_config['source_environment']['source_file']}; "
         f"conda activate {loaded_config['source_environment']['conda_env']}; "
     )
-    config["source_environment"] = src_env
+    # 2.1 - Parse slurm user config account
+    if config["slurm_config"]:
```
LukasNickel, line 211: This will never be true.
Do you mean 
```python
if loaded_config.get("slurm_config")
```
?

garciagenrique: you're completely right @LukasNickel ! 
it was `loaded_config` indeed (thanks tests too... yes, pending thing to do...)

LukasNickel: Use `.get()` if this is optional in the config. If you want to enforce it, adapt the test (Which should probably be done anyway)

garciagenrique: Thanks @LukasNickel ;-)

vuillaut:
Rather than translating slurm options one by one, wouldn't it be easier to allow user to pass any option to the slurm command transparently?

Also, you can already pass these options through environment variables:
https://slurm.schedmd.com/sbatch.html#SECTION_INPUT-ENVIRONMENT-VARIABLES

garciagenrique:
> Rather than translating slurm options one by one, wouldn't it be easier to allow user to pass any option to the slurm command transparently?
> 
> Also, you can already pass these options through environment variables: https://slurm.schedmd.com/sbatch.html#SECTION_INPUT-ENVIRONMENT-VARIABLES

Thanks for sharing this info @vuillaut. I agree that passing vars to slurm should be more easily configurable. However, there are some vars that I would suggest not to allow to be defined by users (error files, arrays, queues for specific particles etc...). 
Can be interesting to add this functionality but I don't see many others to be added. Remember that they should be slurm options in common for all the stages, not stage-specifically.

LukasNickel:
> Rather than translating slurm options one by one, wouldn't it be easier to allow user to pass any option to the slurm command transparently?
> 
> Also, you can already pass these options through environment variables: https://slurm.schedmd.com/sbatch.html#SECTION_INPUT-ENVIRONMENT-VARIABLES

This would be great if we would pass the currently active enviroment directly to the commands, so that you dont need to specify an enviroment in the config (Which I would prefer anyway)

garciagenrique:
> This would be great if we would pass the currently active enviroment directly to the commands, so that you dont need to specify an enviroment in the config (Which I would prefer anyway)

How can we do this ? I'm probably not being able to see what you have in mind :)

PR #158, opened 2022-02-15
Move to lstchainv091

garciagenrique:


LukasNickel:
Did you run a full processing with these?

maxnoe:
There were some changes to the command line arguments of the scripts, did you check if you need to adapt anything because of that?

PR #163, opened 2022-02-18
reorganize dl2_to_sens script to allow  importing

vuillaut:


garciagenrique:
You are breaking the entry point.

Maybe you wanna do something like

```
def dl2_to_sensitivity():
 ...

def main():
   args. ...
   args = ...
   dl2_to_sensitivity(all the args)

if __name__ == "__main__":
   main()
```

garciagenrique (approved):
Everything working fine !

PR #165, opened 2022-02-18
Add Sensitivity ratio plot

vuillaut:


LukasNickel:
These are not used as of now. Do you plan to add them to the sensitivity script?
Could you remove the imports as well so that the CI is happy?

vuillaut:
> These are not used as of now. Do you plan to add them to the sensitivity script?

Yes. The plan is to have a systematic comparison to a selected reference production.
I'll implement that later.

LukasNickel (approved):

Issue #160, opened 2022-02-16
Need to update lstchain config file for DL1AB processing


chaimain:
The lstchain config file https://github.com/cta-observatory/lstmcpipe/blob/master/lstmcpipe/standard_configs/lstchain_standard_config_v091.json does not include the MC tuning (PSF and NSB) information as written in https://github.com/cta-observatory/cta-lstchain/blob/master/lstchain/data/lstchain_dl1ab_tune_MC_to_Crab_config.json.

This is essential as this is THE information required for DL1AB processing.

maxnoe:
But that is only for the specific crab data from November 2021. It's not valid for most data. 

Using it for everything is not possible and the discussion how to provide proper models for all data only started in the last analysis call.


So just including the crab config is I think not the way to go.

chaimain:
Yes. We do not need to tune the NSB based on Crab for general purpose, but for PSF tuning, we can do that, right?

@moralejo mentioned that we can have the NSB tuning turned to False, and then proceed with PSF tuning

chaimain:
For NSB tuning, we can probably have 1 set for Crab at least for now. I am not sure, how we will define 'standard config', as it is part of a larger discussion.

moralejo:
Yes, we have to decide what to do regarding NSB tuning, especially once we have the new MC.

As of now, using the same NSB for all (be it dark or Crab) will anyway be less of a problem than using a single pointing for data taken with different pointings. Using Crab's NSB has the advantage that we get the best possible MC for the analysis of our standard candle - of course, not for the rest of sources.

PSF is a different story, because the default one in MC is too good, so I would definitely apply the PSF tuning to the MC for the analysis of any of the LST data we have taken so far.

PR #168, opened 2022-02-21
Add noise fine tuning

LukasNickel:
Add config keys that trigger the use of `lstchain_tune_nsb` in order to get the image modifier settings for a given observation run.
That should eliminate one step got getting matching monte carlo

### DIFF
--- a/lstmcpipe/config_MC_prod.yml
+++ b/lstmcpipe/config_MC_prod.yml
@@ -18,33 +18,23 @@ workflow_kind: lstchain  # Not expected to be changed by the user
 base_path: /fefs/aswg/workspace/MC_data_simlink  # Default user path
   # lstanalyzer user : /fefs/aswg/data/mc
   #
-# Only for dl1 reprocessing
-#dl1_reference_id: #old id including versions (full directory name)
 prod_type: prod5
-  # Choose between {prod5, prod3}
 obs_date: 20200629_prod5_trans_80
-  # prod5 : `20200629_prod5` or `20200629_prod5_trans_80`
-  # pro3 : `20190415`
-zenith: zenith_20deg  # Taken into account only when `prod_type: prod5` [ thus `obs_date` : 20200629_prod5(_trans_80) ]
+zenith: zenith_20deg
 pointing: south_pointing
-
-# If the image modifier settings should be calculated on the fly
-# dl1_noise_tune_data_run: /fefs/aswg/data/real/DL1/20201121/v0.9.1/tailcut84/dl1_LST-1.Run02988.0000.h5
-# dl1_noise_tune_mc_run: /fefs/aswg/data/mc/DL0/20200629_prod5_trans_80/proton/zenith_20deg/south_pointing/proton_20deg_180deg_run1___cta-prod5-lapalma_4LSTs_MAGIC_desert-2158m_mono.simtel.gz
 
 particles:
   - electron
   - gamma
   - gamma-diffuse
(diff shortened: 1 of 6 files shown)


LukasNickel:
Connected issue: #160

garciagenrique:
Thanks for the PR @LukasNickel, looks fine to me, just a silly suggestion.

lstmcpipe/config_MC_prod.yml
```diff
@@ -27,6 +27,11 @@ obs_date: 20200629_prod5_trans_80
   # pro3 : `20190415`
 zenith: zenith_20deg  # Taken into account only when `prod_type: prod5` [ thus `obs_date` : 20200629_prod5(_trans_80) ]
 pointing: south_pointing
+
+# If the image modifier settings should be calculated on the fly
```
garciagenrique, line 31: The image modifier settings will only work when running a `dl1ab` stage, so I would put `dl1_reference_id`, `dl1_noise_tune_data_run`, `dl1_noise_tune_mc_run` close in the config file.

garciagenrique:
Now that I'm thinking it twice, what would you say about creating 2 `config_MC_prod.yml` files ?
 - One shows how to trigger a "normal" MC prod (starting from `r0_dl1`) - and named `config_MC_prod.yml`
 - The second one shows how to trigger a MC prod starting from dl1ab stage, show (commented) examples of all dl1ab options, i,.e., `dl1_reference_id`, 'dl1_noise_tune_data_run' and `dl1_noise_tune_mc_run`, as well as the `dl1ab` added into the `stages_to_be_run`

LukasNickel:
> 

Good idea. Maybe we should start separating both things

garciagenrique:
> Good idea. Maybe we should start separating both things

Could you then take care of this in this PR, @LukasNickel ? (no rush :) ! )

lstmcpipe/config_MC_prod.yml
```diff
@@ -18,33 +18,23 @@ workflow_kind: lstchain  # Not expected to be changed by the user
 base_path: /fefs/aswg/workspace/MC_data_simlink  # Default user path
   # lstanalyzer user : /fefs/aswg/data/mc
   #
-# Only for dl1 reprocessing
-#dl1_reference_id: #old id including versions (full directory name)
 prod_type: prod5
-  # Choose between {prod5, prod3}
 obs_date: 20200629_prod5_trans_80
-  # prod5 : `20200629_prod5` or `20200629_prod5_trans_80`
```
garciagenrique, line 26: Sorry for being picky here. 
I would suggest to keep some (a less as possible) comments indicating the possible values of certain fields  (users usually are not aware of how many MC simtel prods are). 

Maybe `prod_type` and `obs_date` ? What do you think ? Not sure if anybody would like to run old prods, indeed...

LukasNickel: But why would anyone run anything else than these values?
Its not like we have 200 simulations currently

LukasNickel: But maybe obs date is not self-explanatory indeed

LukasNickel: If they dont know the productions they should stick to the defaults though I think

garciagenrique: Fair enough ! you convinced me :)

garciagenrique (approved):

PR #169, opened 2022-02-23
Add  templates for new production configs

vuillaut:
The proposal here is to create a new workflow allowing LST members to ask for a new production to be run through pull requests.

In this workflow, they open a new pull request following the provided template, the CI will check that the configuration is technically correct, we review that the proposal makes sense and launch the new production once the PR is accepted.
This will simplify the process, avoid mistakes in configs and introduce a better tracking of the requested official production, also allowing all analysers to see what configs and production are discussed or done.


.github/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+
+<!---  
+New Prod Config template
+Add date and prodid and fill the template
+Your Pull Request must include the following files placed in directory `production_configs/yyyymmdd_prod_id`:
+- lstmcpipe config file
+- lstchain config
+- readme.md
+
+Add label `new_prod_config`
```
garciagenrique, line 15: "Add github label `new_prod_config`" could be a bit more explicative maybe ?

garciagenrique:
Thanks a lot @vuillaut for this implementation.

How possible would be to compare (when doing the PR, so running an specific CI only for PRs), that makes a diff between the standard lstchain config file and the proposed one ?

garciagenrique:
> Interesting! But we still launch it by hand if I understand correctly? Its "just" proposals, right?

Exactly. This way we will have a place to record the requested prods from people of the collab, and where other can have a look to check was already launched. Also, this will be another way to check that the configuration used is the correct one / desired by the users (lstchain config starts to get very big and with MANY parameters).

vuillaut:
Let's implement this !

PR #174, opened 2022-02-28
codemeta downloadUrl

vuillaut:


codemeta.json
```diff
@@ -5,7 +5,7 @@
     "codeRepository": "git+https://github.com/cta-observatory/lstmcpipe.git",
     "dateCreated": "2019-09-09",
     "datePublished": "2021-02-18",
-    "downloadUrl": "https://github.com/cta-observatory/lstmcpipe/archive/refs/tags/v0.5.1.tar.gz",
+    "downloadUrl": "https://github.com/cta-observatory/lstmcpipe/archive/master.tar.gz",
```
garciagenrique, line 8: Should it be pointing to master or to the latest stable version ?

vuillaut: latest version would be nice but since we always forget to update it...

garciagenrique (changes requested):
Please refer to latest release (v0.6.0), and next time we will pay attention to update this field too.

codemeta.json
vuillaut: Honorable thought, I just don't see it happening, and more importantly I don't want to increase human maintenance cost.

vuillaut:
Will be updated for the next release.

PR #175, opened 2022-03-01
Add a new production config for tuned PSF, and for zenith 20 deg, south pointing MC

chaimain:
This will be helpful for analyzing sources like some AGNs, which do not require tuning the NSB, but only PSF.


production_configs/20220301_psf_tuned_zen20/lstmcpipe_config.yml
```diff
@@ -0,0 +1,58 @@
...
+  - gamma
+  - gamma-diffuse
+  - proton
+offset_gammas:  # Taken into account only when `prod_type: prod5` [ thus `obs_date`: 20200629_prod5(_trans_80) ]
+  - off0.0deg
+  - off0.4deg
+
+source_environment:
+  source_file: /fefs/aswg/software/conda/etc/profile.d/conda.sh
+  conda_env: lstchain-v0.9.2
```
LukasNickel, line 41: Why not use 0.9.3 directly?

production_configs/20220301_psf_tuned_zen20/lstchain_config.json
```diff
@@ -0,0 +1,275 @@
...
+        }
+      }
+    },
+    "waveform_nsb_tuning":{
+      "nsb_tuning": false,
+      "nsb_tuning_ratio": 0.52,
+      "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
+    },
+    "image_modifier": {
+      "increase_nsb": false,
```
LukasNickel, line 266: This skips the nsb tuning, but I would still set the values to 0 to avoid confusion.

LukasNickel (changes requested):

production_configs/20220301_psf_tuned_zen20/lstmcpipe_config.yml
chaimain: Yes. Indeed

production_configs/20220301_psf_tuned_zen20/lstmcpipe_config.yml
```diff
@@ -0,0 +1,58 @@
...
+  # user example: "source /fefs/home/enrique.garcia/.bashrc; conda activate lst-dev" should be writen:
+  #   source_file: /fefs/home/enrique.garcia/.bashrc
+  #   conda_env: lst-dev
+
+stages_to_be_run:
+  - r0_to_dl1
```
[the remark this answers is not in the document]
chaimain: Ah yes! Forgot to notice that change.

production_configs/20220301_psf_tuned_zen20/readme.md
```diff
@@ -0,0 +1,11 @@
+# New Prod Config 
+
+## 20220301_psf_tuned_zen20
```
LukasNickel, line 3: Thats not what the folder will be named in the end.
Seeing this I think we could improve the PR template a bit. You are the first one to submit, so you notice the initial trouble :)

LukasNickel (changes requested):
@vuillaut @garciagenrique Do you think the first h2 in the readme is useful? I guess its supposed to be the prod_id but its not really apparent to users how thats constructed and including a hardcoded date leads to issues whenever we dont approve at the day the PR was opened.

production_configs/20220301_psf_tuned_zen20/lstmcpipe_config.yml
```diff
@@ -0,0 +1,63 @@
...
+#   $BASE_PATH_DL0/$STAGE/$OBS_DATE/$PARTICLE/$ZENITH/$POINTING/$PROD_ID
+#
+#
+#  This way every MC production will follow the same path logic;
+#    ex;  $BASE_PATH_DL0/{DL0, DL1, DL2, running_analysis}/20200629_prod5/{electron, gamma, gamma-diffuse, proton}/
+#          /zenith_20deg/south_pointing/$PROD_ID
+
+
+workflow_kind: lstchain  # Not expected to be changed by the user
+#prod_id:                 # ex: local_no_n_islands. Default; v00 (if key left empty or None)
```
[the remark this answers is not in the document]
chaimain: Ok

chaimain:
I am also confused with the date to choose as part of the production ID. In the end, it is not when the production begins, and users cannot predict that either. Maybe lstmcpipe can add the date of the start of production by itself and users just provide string additions to the full id?

LukasNickel:
> I am also confused with the date to choose as part of the production ID. In the end, it is not when the production begins, and users cannot predict that either. Maybe lstmcpipe can add the date of the start of production by itself and users just provide string additions to the full id?

It does that internally. There is always
```python
f"{year}{month}{day}_v{lstchain.__version__}"
```
in front and then the monte carlo identifier (prod3/5/trans_80)
I agree that our use of the term `prod_id` in code and config is confusing.

In the config there should just be the suffix describing the settings so `dl1ab_tuned_psf` (or smth like that) in this case.
It will expand to
```
20220302_v0.9.3_prod5_trans_80_dl1ab_tuned_psf
```
(if launched today).
The question is what should be listed in the README and I dont think the final prod_id/directory name is super useful since thats constructed in the code and can easily deviate if we are not super careful

I would propose just to wait what @garciagenrique or @vuillaut  have to say about the matter.

production_configs/20220301_psf_tuned_zen20/readme.md
garciagenrique: And thanks for this @chaimain,
Could you please update your PR having a look again to the template, please ?

garciagenrique: I see another "notation problem".
I think that `psf_tuned_zen20` is ok for the sub-title, the prod_id, as @LukasNickel suggested above, is more explicative if it's `dl1ab_tuned_psf`

chaimain: Ok. Will change it to that

garciagenrique (approved):

PR #176, opened 2022-03-01
Add a new production config for tuned PSF, and for zenith 40 deg, south pointing MC

chaimain:
This will be helpful for analyzing sources like some AGNs, which do not require tuning the NSB, but only PSF. Similar to #175


LukasNickel:
See discussion in the other PR

garciagenrique:
Triggering the discussion here because it was the first PR to be opened.

Would it be interesting to runs a "standard_zenith40deg_180deg" prod (so local peak integrator and tailcuts 8,4), and from this prod re-process dl1ab stage with all the zenith40 deg user prods ? @chaimain @misabelber @jsitarek Or the standard one won't be really necessary ?

production_configs/psf_tuned_zen40/lstmcpipe_config.yml
```diff
@@ -0,0 +1,44 @@
...
+
+
+workflow_kind: lstchain  # Not expected to be changed by the user
+prod_id: zen40az180_dl1ab_tuned_psf        # ex: local_no_n_islands. Default; v00 (if key left empty or None)
+
+base_path: /fefs/aswg/data/mc  # Path for centralised LST MC prods
+
+# Only for dl1 reprocessing - you will also need to change:
+# `r0_to_dl1` for `dl1ab` in "stages_to_be_run"
+dl1_reference_id: 20220215_v0.9.1_prod5_trans_80_local_tailcut_8_4
```
garciagenrique, line 13: The `dl1_reference_id` should be: `20220304_v0.9.3_prod5_trans_80_zen40az180_local_tailcut_8_4`

There wasn't any zen40deg MC prod analysed with v>0.8 up to now, and `20220215_v0.9.1_prod5_trans_80_local_tailcut_8_4` was referring a zen20deg prod (it wasn't written on the prod-id, no we will start to include it)

production_configs/psf_tuned_zen40/lstmcpipe_config.yml
```diff
@@ -0,0 +1,44 @@
+## MC PRODUCTION FULL PIPELINE CONFIGURATION ##
+#
+# Template example to request a MC prod.
+
+
+workflow_kind: lstchain  # Not expected to be changed by the user
+prod_id: zen40az180_dl1ab_tuned_psf        # ex: local_no_n_islands. Default; v00 (if key left empty or None)
```
garciagenrique, line 7: 👍  thanks @chaimain !

LukasNickel:
> Would it be interesting to runs a "standard_zenith40deg_180deg" prod (so local peak integrator and tailcuts 8,4), and from this prod re-process dl1ab stage with all the zenith40 deg user prods ? @chaimain @misabelber @jsitarek Or the standard one won't be really necessary ?

I mean we need a reference production for dl1ab in any case.
Or do you mean a second one with different settings?

garciagenrique:
> Edit: Ah I think I am late on this one. We have one dating on 2022-03-04 now

Exactly. The 40deg reference one is currently running

PR #178, opened 2022-03-02
Clean mc prod template

garciagenrique:
Clean  (try to clarify) `lstmcpipe.config` - from where (i think) people are taking info to create MC prod requests.


README.rst
```diff
@@ -23,14 +23,15 @@ Lukas Nickel, lukas.nickel [at] tu-dortmund.de
...
 To do so, please:
 
-#. open a pull request in lstMCpipe, adding the desired configuration in a new directory named `date_ProdID` in `production_configs`
-#. you may have a look at the `production_configs/20220222_template/` as an example
-#. Add the date and a descriptive production ID name to you directory and config
-#. the requested config must contain:
+#. Check that you are in the `lst-dev` GitHub group (otherwise, please contact us).
+#. Open a pull request in lstMCpipe, adding the desired configuration in a new directory named `date_ProdID` in `production_configs`.
+#. You may have a look at the `production_configs/20220222_template/` as an example.
+#. Add the date and a descriptive production ID name to you directory and configuration.
```
LukasNickel, line 33: Imo thats not a good idea.
lstmcpipe will always take the date the production is launched which can be different from when the PR is opened (or even merged)

garciagenrique: I think that the `date_` is just to keep track in an ordered way inside the package of all the requests

vuillaut: > please contact us

That would not be very useful, change please.

> lstmcpipe will always take the date the production is launched which can be different from when the PR is opened (or even merged)

I agree @LukasNickel that the production date would be prefered.
Although, I am not sure what you mean by "lstmcpipe will always take the date" - the date is only an indication here, as much as the prodID - it's not used anywhere in the code as a source of truth?

LukasNickel: Ok so date of request.
But we should note that this is different from the prod id.
It also doesnt need to take a form of 20220222_template_prod then

LukasNickel: > I agree @LukasNickel that the production date would be prefered. Although, I am not sure what you mean by "lstmcpipe will always take the date" - the date is only an indication here, as much as the prodID - it's not used anywhere in the code as a source of truth?

I mean that something like here:
https://github.com/cta-observatory/lstmcpipe/blob/dc7f205b67c8b0d0df2ab669928e271224a9f92e/production_configs/20220301_psf_tuned_zen20/readme.md?plain=1#L3
is deceiving because you cant actually find the files under that name afterwards but it looks like a prod id

vuillaut: > #. Check that you are in the `lst-dev` GitHub group (otherwise, please contact us).

Moreover, people can still open a PR from their fork if they are not in the group. We don't use GitHub secrets or LST data in lstmcpipe, so I am not sure to understand the requirement to be in the lst-dev group.

vuillaut: > 



> I mean that something like here:
> 
> https://github.com/cta-observatory/lstmcpipe/blob/dc7f205b67c8b0d0df2ab669928e271224a9f92e/production_configs/20220301_psf_tuned_zen20/readme.md?plain=1#L3
> 
> 
> is deceiving because you cant actually find the files under that name afterwards but it looks like a prod id


Ho, I think I understand now.     
The date that is written in the directory structure is set automatically by lstmcpipe as `today` when launched?

If so, you are entirely right, this will be deceiving!
Any proposal?

vuillaut: I'd like to build an index later, but the date will be easy enough to retrieve, so there is no strong requirement there.
I think you are right, it might be better to remove it rather than having something deceiving to people down the road.

garciagenrique: Personally, rather than managing people individually (yes, we don't have secrets and whatever), it's easier a group. That indeed includes lstchian (and not sure if other libraries, but could be).

garciagenrique: My proposal, take out the `date_prodID` and just keep `prodID`. Also the date the PR is open won't be the same (most probable) as the date the prod will be launched. 
But this will also be deceiving.... as the full prod id is set as follows
${date(today)}_v${lstchain-version}_${OBS_date, i.e., kind of prod}_${prod_id_tag}

garciagenrique:
Indeed, 
We could index by "source name" and position, something like 
"RS Oph. Zenith 20deg, South pointing"
"AGNs 40 deg, NSB tunning"

we don't care about when this MC prod was requested, neither when it was launched indeed. 

Inside the `production_confs` directory it will be just the confs by source name and pointing

chaimain:
> Indeed, We could index by "source name" and position, something like "RS Oph. Zenith 20deg, South pointing" "AGNs 40 deg, NSB tunning"
> 
> we don't care about when this MC prod was requested, neither when it was launched indeed.
> 
> Inside the `production_confs` directory it will be just the confs by source name and pointing

I don't think we need to have source names in the MC production. Just having sky position and tuning information will suffice. Maybe in the readme description, it can be extra information on the usage of the particular production.

README.rst
vuillaut: > Personally, rather than managing people individually (yes, we don't have secrets and whatever), it's easier a group. That indeed includes lstchian (and not sure if other libraries, but could be).

but 
- we don't need to manage anything to allow people opening a PR here
- and we can't manage the lst-dev group anyway

vuillaut:
> Indeed, We could index by "source name" and position, something like "RS Oph. Zenith 20deg, South pointing" "AGNs 40 deg, NSB tunning"
> 
> we don't care about when this MC prod was requested, neither when it was launched indeed.
> 
> Inside the `production_confs` directory it will be just the confs by source name and pointing

What do you mean by index?

vuillaut (approved):

README.rst
vuillaut: So, I changed the instructions in this PR to require only a prod ID with no date.

PR #185, opened 2022-03-03
Add new production for 40zd , 180deg pointing and tuned NSB

misabelber:
# NSB tuned Config                                                                                  
                                                                                                    
Config for 40zd and 180deg with tuning to the NSB of the required field                             
                                                                                                    
## Why this config is needed                                                                        
                                                                                                    
Processing of data with specific NSB


garciagenrique:
Hi @misabelber,

I see that the configs between this PR and the one that @jurysek opened differs only at `image_modifier` parameters (same as the one opened at 40deg by @chaimain). This was the reason why we wanted to originally put the source name within the prod-id, to clarify or ease the config comparison between the requested MC prods (we didn't realised that the repo is public indeed).

I ping @vuillaut to think in a good way of solving this issue and ease the configs comparison rather than doing a `diff ...`.
An option could be tagging the different sources as `source1`, `source2`, ... and later in a private repo/wiki page indicate the real name of the source. 

Summing up. If you don't mind, once we have found a solution, I will be directly pushing to this PR to create a valid `lstmcpipe_config` (the lstchain one looks perfectly fine!)

PR #186, opened 2022-03-03
Do no specify slurm account in config (default to the one that the user belongs to)

garciagenrique:


vuillaut (changes requested):
change in template as well ?

garciagenrique:
> change in template as well ?

Yes, this way we don't need to change it us later (user MC prods are gonna be launched from this user so that they will be centralised).

morcuended:
Why not leave it by default to whichever account the user belongs to?

garciagenrique:
> Why not leave it by default to whichever account the user belongs to? (without the need of explicitly specifying it)

You are right. Even more that this arg is optional

PR #187, opened 2022-03-04
change all partitions to long

garciagenrique:
Computing time sometimes depends on the machine.
electrons and point like gammas should be running on short, but sometimes they run on timeout at the cluster


LukasNickel:
Maybe its a priority/utilization thing as well?
Anyway I dont think we hurt anyone by using the long queue, lets just go for it

LukasNickel (approved):

PR #188, opened 2022-03-05
Producing source dependent MC production for zenith 20 deg

chaimain:
Starting source-dependent MC production, after which we can produce with some MC tuning in a separate PR

### DIFF
--- a/production_configs/srcdep_zen20/lstchain_config.json
+++ b/production_configs/srcdep_zen20/lstchain_config.json
@@ -1,268 +1,260 @@
 {
-    "source_config" : {
-      "EventSource": {
-        "allowed_tels": [1],
-        "max_events": null
+  "source_config" : {
+    "EventSource": {
+      "allowed_tels": [1],
+      "max_events": null
+    },
+    "LSTEventSource": {
+      "default_trigger_type": "ucts",
+      "allowed_tels": [1],
+      "min_flatfield_adc": 3000,
+      "min_flatfield_pixel_fraction": 0.8,
+      "calibrate_flatfields_and_pedestals": false,
+      "EventTimeCalculator": {
+        "dragon_reference_counter": null,
+        "dragon_reference_time": null
       },
-      "LSTEventSource": {
-        "default_trigger_type": "ucts",
-        "allowed_tels": [1],
-        "min_flatfield_adc": 3000,
-        "min_flatfield_pixel_fraction": 0.8,
-        "calibrate_flatfields_and_pedestals": false,
-        "EventTimeCalculator": {
-          "dragon_reference_counter": null,
-          "dragon_reference_time": null
-        },
-        "PointingSource":{
-          "drive_report_path": null
-        },
-        "LSTR0Corrections":{
-          "calib_scale_high_gain":1.088,
-          "calib_scale_low_gain":1.004,
-          "drs4_pedestal_path": null,
-          "calibration_path": null,
-          "drs4_time_calibration_path": null
-        }
-      }
-    },
-  
-    "events_filters": {
-      "intensity": [0, Infinity],
-      "width": [0, Infinity],
-      "length": [0, Infinity],
-      "wl": [0, Infinity],
-      "r": [0, Infinity],
-      "leakage_intensity_width_2": [0, Infinity]
-    },
-  
-    "tailcut": {
-      "picture_thresh":8,
-      "boundary_thresh":4,
-      "keep_isolated_pixels":false,
-      "min_number_picture_neighbors":2,
-      "use_only_main_island":false,
-      "delta_time": 2
-    },
-    "tailcuts_clean_with_pedestal_threshold": {
-      "picture_thresh":8,
-      "boundary_thresh":4,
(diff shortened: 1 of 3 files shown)


chaimain:
@SeiyaNozaki Can you just confirm if the parameters are correct as we discussed?

production_configs/srcdep_zen20/lstchain_config.json
```diff
@@ -0,0 +1,275 @@
...
+      "width",
+      "length",
+      "x",
+      "y",
+      "wl",
+      "skewness",
+      "kurtosis",
+      "time_gradient",
+      "leakage_intensity_width_2"
+    ],
```
SeiyaNozaki, line 148: you need to use parameters described in `lstchain_src_dep_config.json`.

production_configs/srcdep_zen20/lstchain_config.json
```diff
@@ -0,0 +1,275 @@
...
+      "y",
+      "wl",
+      "skewness",
+      "kurtosis",
+      "time_gradient",
+      "leakage_intensity_width_2",
+      "log_reco_energy",
+      "reco_disp_norm",
+      "reco_disp_sign"
+    ],
```
SeiyaNozaki, line 188: same above

production_configs/srcdep_zen20/lstchain_config.json
```diff
@@ -0,0 +1,275 @@
...
+    "disp_classification_features": [
+      "log_intensity",
+      "width",
+      "length",
+      "wl",
+      "skewness",
+      "kurtosis",
+      "time_gradient",
+      "leakage_intensity_width_2"
+    ],
```
SeiyaNozaki, line 172: same above (though disp reconstruction is not important for source-dependent analysis)

SeiyaNozaki (commented):
Thank you for this request @chaimain !
You need to use input parameters for source-dependent analysis at the training step. 
The other concerns are:
- For source-dep analysis, point gamma MC should be used for RF training. I'm not sure where it can be configured but should be used.
- when creating source-dep IRF, we need to use an option of `--source-dep`. Can this option used just by modifying the configuration file?
- I guess dl2_to_sensitivity will not work with source-dep analysis

production_configs/srcdep_zen20/lstchain_config.json
chaimain: I used the standard `lstchain_src_dep_config.json` from `lstchain/data/` but it seems to not completely follow the `lstchain_standard_config.json`, mainly on the calibration part. Can you check again? I tried to update with what I think is relevant here from both configs.

chaimain:
> Thank you for this request @chaimain ! You need to use input parameters for source-dependent analysis at the training step. 

Done, thanks for checking.
The other concerns are:
> 
> * For source-dep analysis, point gamma MC should be used for RF training. I'm not sure where it can be configured but should be used.

I have removed diffuse-gamma MC from the workflow. Will that be ok, or do you want it to be only restricted in the RF training?

> * when creating source-dep IRF, we need to use an option of `--source-dep`. Can this option used just by modifying the configuration file?

I will raise an issue here to add such support.

> * I guess dl2_to_sensitivity will not work with source-dep analysis

For this, I have commented it from the lstmcpipe stages to be run.

production_configs/srcdep_zen20/lstchain_config.json
```diff
@@ -0,0 +1,267 @@
...
+    "nsb_tuning_ratio": 0.52,
+    "spe_location": "lstchain/data/SinglePhE_ResponseInPhE_expo2Gaus.dat"
+  },
+  "image_modifier": {
+      "increase_nsb": false,
+      "extra_noise_in_dim_pixels": 0,
+      "extra_bias_in_dim_pixels": 0,
+      "transition_charge": 0,
+      "extra_noise_in_bright_pixels": 0,
+      "increase_psf": false,
```
LukasNickel, line 263: Are you sure you want this to be False?
Just making sure :)

chaimain: Yes, because, we should have some standard source-dependent DL1 files as a reference for the tuning, as we did earlier for zenith 40 deg, source independent MC production.

So, these 2 PRs are the primer for another set of PRs for 
1. Source-dependent, PSF tuning, zenith 20 and 40 deg
2. Source-dependent, PSF + NSB tuning, zenith 20 and 40 deg (depending on the selected NSB)

chaimain: I am not sure how much is too much for MC production while we wait for the all-sky MC set ;-)

LukasNickel: Ah right thats a new base, not dl1ab.
My bad

LukasNickel: > I am not sure how much is too much for MC production while we wait for the all-sky MC set ;-)

Guess we have some things to discuss in the call

chaimain:
We can probably close this one and the 40 deg one. Unless anyone has any objections.

PR #192, opened 2022-03-08
create prod_logs directory outside package. 

garciagenrique:
Every launched prod will save log, slurm and config files to this dir.

`create_log_files` and `update_scancel_file` moved from `workflow_management` to `io` submodule.

### DIFF
--- a/lstmcpipe/__init__.py
+++ b/lstmcpipe/__init__.py
@@ -1,1 +1,5 @@
+from pathlib import Path
+
+lstmcpipe_root_dir = Path(__file__).absolute().parents[1]
+
 __version__ = "0.6.1"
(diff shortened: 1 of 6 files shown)


vuillaut:
The idea is not to save the logs in this repo, right?

garciagenrique:
> The idea is not to save the logs in this repo, right?

Not remotely, yes locally. 
`prod_logs` has been added to the .gitignore, and every time a prod is launched, the logs are saved within this directory. Therefore all the prods that will be triggered (from LP, let's say) will stay inside the `lstmcpipe` package, but won't be added to the repo.

vuillaut (approved):
lgtm

PR #194, opened 2022-03-10
Documentation page with produced configs

vuillaut:
This produce a doc page with the list of produced configs


docs/produce_config_page.py
```diff
@@ -0,0 +1,64 @@
...
+
+    prod_txt += tabulate(prod_list, ['Request date', 'Directory name', 'Prod ID'], tablefmt='rst')
+
+    with open(prod_file, 'a') as f:
+        f.write(prod_txt)
+
+
+
+if __name__ == '__main__':
+    root_dir = Path(__file__).parent.joinpath('..').resolve()
```
garciagenrique, line 59: the `__init__.py` inside lstmcpipe contains a root_dir pointer :)

docs/produce_config_page.py
```diff
@@ -0,0 +1,64 @@
...
+    prod_dir = root_dir.joinpath('production_configs')
+
+    lstmcpipe_repo_prod_config_url = 'https://github.com/cta-observatory/lstmcpipe/tree/master/production_configs/'
```
garciagenrique, line 62: `lstmcpipe_repo_prod_config_url` is also a var inside the `add_prod_table` function, but you don't pass it as argument. 
Is this really working ?

vuillaut: it's working but bad practice, I will change

docs/produce_config_page.py
vuillaut: Hum, after thinking about it, it would add lstmcpipe to the requirements install, which is not really necessary just for that

garciagenrique (approved):
Thanks for the automatisation !

PR #195, opened 2022-03-10
Former configs added to production_configs

vuillaut:
This PR makes sense after #194


production_configs/20210416_v0.7.3_prod5_trans_80_local_taicut_8_4/log_onsite_mc_r0_to_dl3_20210416_v0.7.3_prod5_trans_80_local_taicut_8_4.yml
```diff
@@ -0,0 +1,2438 @@
+r0_to_dl1:
```
garciagenrique, line 1: This are the full logs for a single MC prod. I don't think this should go inside the repo. 
Note that there has been some renaming of these filenames, but they should all contain the prod_id within their name.

production_configs/20220218_v0.9.2_prod5_trans_80_dl1ab_tune_MC_to_Crab/log_lstmcpipe0.6.0_20220218_v0.9.2_prod5_trans_80_dl1ab_tune_MC_to_Crab.yml
```diff
@@ -0,0 +1,612 @@
+r0_to_dl1:
```
garciagenrique, line 1: Same comment as above

garciagenrique (changes requested):

production_configs/20210416_v0.7.3_prod5_trans_80_local_taicut_8_4/log_onsite_mc_r0_to_dl3_20210416_v0.7.3_prod5_trans_80_local_taicut_8_4.yml
vuillaut: I must have taken the wrong files then. Will DM you.

vuillaut: Ok I think this is fixed

garciagenrique:
CI failing because `merging_options:  no_image: True` was not implemented until lstmcpipev>=0.7.5. Should I add a default `.get()` for not changing the lstmcpipe configs used ?

vuillaut:
> CI failing because `merging_options: no_image: True` was not implemented until lstmcpipev>=0.7.5. Should I add a default `.get()` for not changing the lstmcpipe configs used ?

I would rather implement the test selection to run only on the requested configs.

garciagenrique:
created issue #199

PR #200, opened 2022-03-16
Validate only added production config

vuillaut:


vuillaut:
After some trials and errors, this works.

See https://github.com/cta-observatory/lstmcpipe/runs/5577799020?check_suite_focus=true 
applied to a test config I now removed from this PR.

LukasNickel (approved):
Nice, this should help reviewing requests!

garciagenrique (approved):

PR #205, opened 2022-03-31
Path config

vuillaut:


vuillaut:
I believe this is ready to be merged in the refactor branch so you can use it @garciagenrique

lstmcpipe/config/tests/test_config.py
```diff
@@ -1,119 +1,6 @@
...
-    assert parsed_config["stages_to_run"] == ["r0_to_dl1"]
-    assert parsed_config["workflow_kind"] == "lstchain"
-
-    new_dummy = config.copy()
-    new_dummy["slurm_config"] = {"user_account": "aswg"}
-    new_parsed_config = parse_config_and_handle_global_vars(new_dummy)
-    assert new_parsed_config["batch_config"]["slurm_account"] == "aswg"
+def test_base_config():
+    cfg = config.base_config()
+    assert "source_environment" in cfg
```
garciagenrique, line 6: Could you add all `workflow_kind` and `prod_id` (and I would say that `slurm config` too

garciagenrique (commented):
Nothing against of changing the yaml lib, but which are the exact needs of this change ?

garciagenrique:
Looks nice otherwise !  Merging to branch in the moment we resolve the `test_config` conversation

lstmcpipe/config/tests/test_config.py
vuillaut: in the test?
sure

vuillaut:
> Nothing against of changing the yaml lib, but which are the exact needs of this change ?

Ha yes good question.
`ruamel.yaml` is much more powerful and is able to handle the order of the loaded config as well as keeping comments, both things that are needed imo.

lstmcpipe/config/tests/test_config.py
vuillaut: Done

garciagenrique (approved):

PR #214, opened 2022-04-08
Test prod script

vuillaut:
Small script to generate a mini prod tree on the cluster with its paths config

### DIFF
--- a/lstmcpipe/scripts/generate_test_lapalma.py
+++ b/lstmcpipe/scripts/generate_test_lapalma.py
@@ -7,30 +7,74 @@ from pathlib import Path
 from tqdm import tqdm
 from datetime import date
 import os
+import argparse
 
 from lstmcpipe.config import paths_config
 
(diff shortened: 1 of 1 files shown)


garciagenrique:
Imo, I don't think this script is needed in the lib.

vuillaut:
We'll need this script in the future, I'd like to save it somewhere. Where should I put it ?

garciagenrique (changes requested):
data_test directory at LP doesn't contain off0.4deg gammas (I added by hand). If we create a test_subsample, please include all particles to check a real (although small) workflow.

vuillaut:
> data_test directory at LP doesn't contain off0.4deg gammas (I added by hand). If we create a test_subsample, please include all particles to check a real (although small) workflow.

Regenerate using the up-to-date script please.

garciagenrique (approved):
Thanks @vuillaut !

PR #216, opened 2022-04-08
Fix stages and config

vuillaut:
Fixes quite some issues with the stages and the config, and how information flows in the different stages.

The pipeline now runs (or seems to) from R0 to IRFs but only 1 DL1 is produced 😅 
So some more work to do...

### DIFF
--- a/lstmcpipe/config/paths_config.py
+++ b/lstmcpipe/config/paths_config.py
@@ -48,7 +48,7 @@ class PathConfig:
         if self.paths == {}:
             raise ValueError("Empty paths, generate first")
 
-        config_to_save['prod_type'] = {self.__class__.__name__}
+        config_to_save['prod_type'] = self.__class__.__name__
         config_to_save['prod_id'] = self.prod_id
         config_to_save['stages_to_run'] = self.stages
         config_to_save['stages'] = self.paths
--- a/lstmcpipe/config/paths_config.py
+++ b/lstmcpipe/config/paths_config.py
@@ -76,20 +76,9 @@ class PathConfigProd5Trans80(PathConfig):
         self.zenith = zenith
         self.base_dir = \
             '/fefs/aswg/data/mc/{data_level}/20200629_prod5_trans_80/{particle}/{zenith}/south_pointing/{prod_id}'
-        self.training_particles = [
-            'gamma-diffuse',
-            'proton'
-        ]
-        self.testing_particles = [
-            'gamma',
-            'electron',
-            'proton',
-            'gamma-diffuse',
-        ]
-        self.point_src_offsets = [
-            'off0.0deg',
-            'off0.4deg'
-        ]
+        self.training_particles = ['gamma-diffuse', 'proton']
+        self.testing_particles = ['gamma', 'electron', 'proton', 'gamma-diffuse']
+        self.point_src_offsets = ['off0.0deg', 'off0.4deg']
         self.paths = {}
         self.stages = [
             'r0_to_dl1',
@@ -98,7 +87,7 @@ class PathConfigProd5Trans80(PathConfig):
             'train_pipe',
             'dl1_to_dl2',
             'dl2_to_sensitivity',
-            'dl2_to_irfs'
+            'dl2_to_irfs',
         ]
 
     @property
@@ -331,12 +320,9 @@ class PathConfigProd5Trans80(PathConfig):
             }
             return d
 
-        for gamma_part in ['gamma-diffuse', 'gamma']:
-            if gamma_part == 'gamma-diffuse':
-                paths.append(path_dict(gamma_part, 'diffuse'))
-            else:
-                for offset in self.point_src_offsets:
-                    paths.append(path_dict(gamma_part, offset))
+        # sensitivity can be computed only on point source gammas at the moment
+        for offset in self.point_src_offsets:
+            paths.append(path_dict('gamma', offset))
 
         return paths
 
--- a/lstmcpipe/config/paths_config.py
+++ b/lstmcpipe/config/paths_config.py
@@ -82,7 +82,8 @@ class PathConfigProd5Trans80(PathConfig):
         self.testing_particles = [
             'gamma',
             'electron',
-            'proton'
+            'proton',
+            'gamma-diffuse',
         ]
         self.point_src_offsets = [
             'off0.0deg',
@@ -267,13 +268,13 @@ class PathConfigProd5Trans80(PathConfig):
 
     @property
     def train_pipe(self):
-        paths = [{
+        paths = {
             'input': {
                 'gamma': self.merge_output_file('gamma-diffuse', 'train'),
                 'proton': self.merge_output_file('proton', 'train')
             },
             'output': self.models_path()
-        }]
+        }
         return paths
 
     @property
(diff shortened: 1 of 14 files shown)


lstmcpipe/config/pipeline_config.py
```diff
@@ -196,7 +191,7 @@ def complete_lstmcpipe_config(loaded_config):
         "PathConfigAllSky": "prod_all_sky",
     }
 
-    suffix_id = "_{}_{}".format(all_mc_prods[prod_type], sufix_prod_id)
+    suffix_id = "_{}".format(suffix_prod_id)
```
garciagenrique, line 194: Why taking out the {`prod5`, `prod5_trans_80`, `prod_all_sky`} from the prod_id ? 
In this case remove the `all_mc_prods` dict too, as it is no longer used.

lstmcpipe/config/pipeline_config.py
```diff
@@ -112,9 +111,6 @@ def config_valid(loaded_config):
             f"Please select an allowed `workflow_kind`: {allowed_workflows}"
         )
 
-    prod_type = loaded_config["prod_type"]
-    if prod_type not in allowed_prods:
```
garciagenrique, line 116: Why shouldn't we check the prod type ?
Same comment. 
If we are not checking which type of prod the config file belongs to, erase the `allowed_prods` list above.

vuillaut: why would we need to check the `prod_type`???

vuillaut: currently `prod_type: prod5`
why would the stages need to know that?

vuillaut: the goal here is to remove the hard-coded parts, not replace them

lstmcpipe/config/pipeline_config.py
vuillaut: I really don't  understand the need for this, other that logging.
But logs contain the config, which has this information.

garciagenrique: Filenaming of individual files and dirs will be shorter, it will be less self explicative. Nobody will go to the logs (that are indeed where exactly ?) to check if a prod is prod5, prod5_trans_80 if it is not written on the dir, imo

garciagenrique: Screenshot 2022-04-10 at 00 17 06

garciagenrique: we need to be consistent with naming of dirs (we create to many in too many different places). To be discussed

vuillaut: Ok for the directory naming that will help people navigating the dir tree.
I have added `prod_type` in the config in #217 . It can be used directly for logging and naming purposes.

lstmcpipe/config/pipeline_config.py
vuillaut: No need to check an allowed list. `prod_type` is just additional info but there is nothing mandatory.

lstmcpipe/config/pipeline_config.py
```diff
@@ -161,12 +162,12 @@ def complete_lstmcpipe_config(loaded_config):
 
         # TODO parse version from hiPeRTA module
         import lstchain
-        base_prod_id = f"{year}{month}{day}_vRTA420_v{lstchain.__version__}"
+        base_prod_id = f"{year}{month}{day}_vRTA420_v{lstchain.__version__}:"
```
garciagenrique, line 165: Why two points and only in the hiperta case ?

lstmcpipe/config/paths_config.py
```diff
@@ -267,13 +269,13 @@ def models_path(self):
 
     @property
     def train_pipe(self):
-        paths = [{
+        paths = {
```
garciagenrique, line 272: I would suggest to return to list also for the train_pipe stage, despite it contains a single item in the list. 
We won't need to change the test in were we check the type of each class attribute ... (see other comment)

lstmcpipe/lstmcpipe_start.py
```diff
@@ -221,7 +221,7 @@ def main():
     if "train_pipe" in stages_to_run:
 
         job_from_train_pipe = batch_train_pipe(
-            lstmcpipe_config,
+            lstmcpipe_config['stages']['train_pipe'],
```
garciagenrique, line 224: (....) and here we will just need to do a `lstmcpipe_config['stages']['train_pipe'][0]`

lstmcpipe/config/pipeline_config.py
vuillaut: typo

lstmcpipe/config/paths_config.py
vuillaut: I only tried to match your implementation of the stages here 😅 
As I understood the current implementation, we could not run several trainings in parallel anyway (because of the jobs dependency handling).
I would be happy if that can be done though.

lstmcpipe/lstmcpipe_start.py
vuillaut: So, no.
If we generate and pass a list here, the assumption will be that everything in that list will be handled, not just the first item. So either the stage `train_pipe` is implemented in a way to handle the complete list (see above comment, not so easy imo), or we pass a single dict - matching the stage logic.

garciagenrique: pass a single dict then, imo

lstmcpipe/config/tests/test_paths_config.py
```diff
@@ -53,73 +54,64 @@ def test_path_config_save():
...
+           f"/fefs/aswg/data/mc/DL1/20200629_prod5_trans_80/gamma/zenith_20deg/south_pointing/{prod_id}/off0.4deg/train"
     assert cfg.test_dir("gamma") == \
-           "/fefs/aswg/data/mc/DL1/20200629_prod5_trans_80/gamma/zenith_20deg/south_pointing/dummy_prodid/off0.4deg/test"
+           f"/fefs/aswg/data/mc/DL1/20200629_prod5_trans_80/gamma/zenith_20deg/south_pointing/{prod_id}/off0.4deg/test"
 
     for stg in cfg.stages:
 
         prop = getattr(cfg, stg)
-        assert isinstance(prop, list)
+        # assert isinstance(prop, list) or isinstance(prop, dict)
```
garciagenrique, line 105: Then, this line can be un-commented, right ?

lstmcpipe/config/tests/test_paths_config.py
```diff
@@ -53,73 +54,64 @@ def test_path_config_save():
...
 
         for path in prop:
 
-            assert isinstance(path, dict)
-            if stg != "dl2_to_irfs":  # TODO and merge_dl1 !!
-                assert all(item in ["input", "output"] for item in path)
-            else:
-                assert all(item in ["input", "output", "options"] for item in path)
+            # assert isinstance(path, dict)
```
garciagenrique, line 109: As well as this one, right ? every item in `getattr` should be a dict always (except for `train_pipe`)

lstmcpipe/config/tests/test_paths_config.py
```diff
@@ -53,73 +54,64 @@ def test_path_config_save():
...
+            # assert all(item in ["input", "output", "options"] for item in path)
```
garciagenrique, line 110: why not to check this one ? (maybe I'm missing something)

vuillaut: fixed

lstmcpipe/config/tests/test_paths_config.py
vuillaut: fixed

PR #217, opened 2022-04-11
write prod_type from class name in the config file

vuillaut:


vuillaut:
Included these changes in #216

PR #234, opened 2022-04-14
autoversioning from git tags

vuillaut:


lstmcpipe/__init__.py
```diff
@@ -1,4 +1,4 @@
 from pathlib import Path
 
 lstmcpipe_root_dir = Path(__file__).absolute().parents[1]
-__version__ = "0.6.1"
```
maxnoe, line 4: you should still import `__version__` here so one can do `from lstmcpipe import __version__`

maxnoe (changes requested):
There are some important steps missing in this. See my PRs in lstchain and ctapipe.

`_dev_version` needs to be a directory, not a python file, to be able to exclude it in wheels. You then need to exclude that package for the`find_package` command in setup.py / setup.cfg.

Also, the `_version.py` should be added to the `.gitignore.

vuillaut:
> There are some important steps missing in this. See my PRs in lstchain and ctapipe.
> 
> `_dev_version` needs to be a directory, not a python file, to be able to exclude it in wheels. You then need to exclude that package for the`find_package` command in setup.py / setup.cfg.
> 
> Also, the `_version.py` should be added to the `.gitignore.

Thank you Max, will do right now

vuillaut:
I must be missing something, the imports fail

vuillaut:
@maxnoe could you review again please?

PR #252, opened 2022-05-06
avoid processing directories not containing simtel files

garciagenrique:


lstmcpipe/io/data_management.py
```diff
@@ -67,40 +68,52 @@ def query_continue(question, default="no"):
         return answer
 
 
-def check_data_path(data_path):
+def check_data_path(data_path, glob=None, rglob=None):
```
vuillaut, line 71: isn't `glob` and `rglob` completely redundant??
one could pass `glob='**/desired_glob'` instead of rglob, no?

lstmcpipe/stages/mc_process_dl1.py
```diff
@@ -173,8 +181,7 @@ def r0_to_dl1(
 
     log.info("Working on DL0 files in {}".format(input_dir))
 
-    check_data_path(input_dir)
-    raw_files_list = [file.resolve().as_posix() for file in Path(input_dir).rglob("**/*.simtel.gz")]
+    raw_files_list = get_input_filelist(rglob_pattern="*.simtel.gz")
```
vuillaut, line 184: simtel extension might change in the near future, see:
https://github.com/cta-observatory/lst-sim-config/issues/42#issuecomment-1104898584

Shall we address that possibility already?

vuillaut (approved):

PR #265, opened 2022-05-10
Check for exit code of r0_to_dl1 

### LINKED ISSUE #260: check for exit code
If not 0, re-run command / job (once).
If still 0, handle output.

vuillaut:
Fixes #260


lstmcpipe/utils.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+import subprocess
+
+
+def rerun_cmd(cmd, outfile, max_ntry=2):
```
garciagenrique, line 5: Don't you want to add test for this, @vuillaut ? 😇

lstmcpipe/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    ntry = 1
+    while ret != 0 and ntry <= max_ntry:
+        ret = subprocess.run(cmd)
+        if ret.returncode != 0:
+            failed_jobs_subdir = Path(outfile.parent).joinpath('failed_outputs')
+            if outfile.exists():
+                failed_jobs_subdir.mkdir(exist_ok=True)
+                outfile_target = failed_jobs_subdir.joinpath(outfile.name)
+                print(f"Move failed output file from {outfile} to {outfile_target}. try #{ntry}")
+                outfile.rename(outfile_target)
```
garciagenrique, line 29: Why not using a `shutil.move` ?

lstmcpipe/scripts/script_batch_filelist_lst.py
```diff
@@ -54,7 +54,8 @@ def main():
             if args.config_file:
                 cmd.append("--config={}".format(args.config_file))
 
-            subprocess.run(cmd)
+            outfile = Path(args.output_dir).joinpath('dl1_' + file.replace('.simtel.gz', '.h5'))
```
garciagenrique, line 57: We are doing a loop over a list of files within the a `*.sublist` file. you can directly recover `outfile` as `file`

garciagenrique (changes requested):

lstmcpipe/utils.py
vuillaut: 👀

lstmcpipe/scripts/script_batch_filelist_lst.py
vuillaut: `file` is the input file here
`rerun_cmd` needs the  output file

lstmcpipe/utils.py
vuillaut: done

lstmcpipe/scripts/script_batch_filelist_lst.py
garciagenrique: ah, right.

garciagenrique: `file` is an str absolute path , though. You would need to do Path(file).name to recover just the file name

vuillaut: thanks, fixed

garciagenrique (approved):

lstmcpipe/scripts/script_batch_filelist_lst.py
```diff
@@ -53,7 +53,9 @@ def main():
             if args.config_file:
                 cmd.append("--config={}".format(args.config_file))
 
-            subprocess.run(cmd)
+            file = Path(file)
```
maxnoe, line 56: Just do `type=Path` in the argumentparse definition above.

lstmcpipe/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    subdir_failures: str
+    outfile: Path
+        path to the cmd output file
+    max_ntry: int
+    run_kwargs: kwargs for subprocess.run
+    """
+    outfile = Path(outfile)
+    ret = -1
+    ntry = 1
+    while ret != 0 and ntry <= max_ntry:
```
maxnoe, line 23: ret will never be == 0, since ret is the `CompletedProcess` instance. You either need to check `ret.returncode` or assign `ret.returncode` to ret.

E.g. like this:

```
ret = CompletedProcess(args=[], returncode=-1)

while ret.returncode !=0 or ntry <= max_ntry:
    ...
```

at the momemnt I think you are always executing the script max_ntry even if it succeeds.

maxnoe (changes requested):

lstmcpipe/scripts/script_batch_filelist_lst.py
vuillaut: `file` is read from the filelist file, not passed through argparse

lstmcpipe/utils.py
vuillaut: Well spotted, thanks. Fixed and unit tested with lstchain as well.

maxnoe (approved):

lstmcpipe/utils.py
maxnoe: I think `rename` does not work across file system boundaries, so using `shutil.move` might be needed for some specific cases:

```
In [2]: p = Path('foo')

In [3]: p.rename('/tmp/bar')
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
Input In [3], in <cell line: 1>()
----> 1 p.rename('/tmp/bar')

File /usr/lib/python3.10/pathlib.py:1232, in Path.rename(self, target)
   1222 def rename(self, target):
   1223     """
   1224     Rename this path to the target path.
   1225 
   (...)
   1230     Returns the new Path instance pointing to the target path.
   1231     """
-> 1232     self._accessor.rename(self, target)
   1233     return self.__class__(target)

OSError: [Errno 18] Invalid cross-device link: 'foo' -> '/tmp/bar'
```

PR #285, opened 2022-05-20
fix DL1ab config generation

vuillaut:


garciagenrique (commented):
I don't see any special dev except the change of name in the vars ? Was it working before ? or is still wip ?

vuillaut:
> I don't see any special dev except the change of name in the vars ? Was it working before ? or is still wip ?

Nothing else to do, the generate script uses `--prod_id` arg and it was not existing for DL1ab classes

garciagenrique (approved):

PR #295, opened 2022-05-23
New production config (src3_dec3476_4822_tuned_nsb)

SeiyaNozaki:
from my fork repository:

# New Prod config

## Prod_ID
src3_dec3476_4822_tuned_nsb

Config for a source with dec 3476/4822 and NSB tuning

## Why this config is needed 
See a quick check on the wiki page (src3):
https://www.lst1.iac.es/wiki/index.php/MC_analysis_and_IRF_production#lstmcpipe_productions


production_configs/20220523_src3_dec3476_4822_tuned_nsb/lstchain_config_2022-05-23.json
```diff
@@ -0,0 +1,315 @@
...
+        "width",
+        "length",
+        "x",
+        "y",
+        "wl",
+        "skewness",
+        "kurtosis",
+        "time_gradient",
+        "leakage_intensity_width_2",
+	"sin_az_tel",
```
vuillaut, line 168: Hi @SeiyaNozaki 
Unfortunately, that parameter is not yet computed in lstchain.
You will have to use `az_tel` - do  you foresee issues with that?

vuillaut (changes requested):

production_configs/20220523_src3_dec3476_4822_tuned_nsb/lstchain_config_2022-05-23.json
SeiyaNozaki: This node contains only north-pointing events, which means azimuth distribution is not continuous (0<AZ<90 and 270<Az<360). So I'm afraid if it will work properly with `az_tel`... For example, I'm not sure how it works when using a node of `theta_32.059_az_355.158` (around the border of az=0/360) as test samples.

For the moment, we want to check if it works properly using the standard input parameters of the current lstchain (without sin(az_sin), signed timed gradient/skewness). Anyway we can compare some results by preparing RF training by ourselves if we have NSB-tuned DL1 files.

So I would propose to use `az_tel` here, and check the behavior later. Is it fine?

vuillaut: Yes if it's ok to test with `az_tel` first, we could launch it today

SeiyaNozaki: we will check processed data and ask for a request again only if needed!

vuillaut: looks good

vuillaut:
🚀

SeiyaNozaki:
it seems proton DL1 files of `node_corsika_theta_7.093_az_31.09_` and `node_corsika_theta_7.093_az_328.91_` were not produced... not only for this NSB tuned one but also standard one (`20220511_allsky_std`). I'm not sure this is a known issue, but just in case...

PR #301, opened 2022-05-24
test valid config

vuillaut:


vuillaut:
config fails, as expected:
https://github.com/cta-observatory/lstmcpipe/runs/6576180962?check_suite_focus=true

.github/workflows/CI.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         with:
           files: |
             production_configs/*/*.yml
```
garciagenrique, line 66: why don't we just use `*.yaml` extension, and therefore we change all the `*.yml` that we are using

vuillaut:
@misabelber 
Your lstchain config was not valid (missing `,`) and the automatic checks unfortunately did not work :-/ 
This PR fixes the config but I will have to restart the prod.

.github/workflows/CI.yml
vuillaut: we do use `.yaml` [as recommended ](https://yaml.org/faq.html)
Better look for both just in case...

misabelber:
Oops, I'm so sorry for this! I should have checked somehow  before if the config file was properly edited... Thank you very much for fixing it.

PR #306, opened 2022-05-25
Running from dl0 to IRFs for dec_4822

giorgio-pirola:
# New Prod config

PR #310, opened 2022-05-27
testing diffuse gamma DL1 production

jurysek:
# New Prod config

## Prod_ID
20220527_src2_diffgamma

## Why this config is needed 
R0->DL1 processing for the new testing diffuse gammas at four nodes along the src2 path.


jurysek:
@vuillaut CI failed with 'Missing: workflow_kind key', but the key is present in the config file. Do you have an idea what could be wrong?

vuillaut:
Could you please merge the master branch? There was an issue with protobuf messing with the install and unit tests.
The config should pass the tests also,  let's see..

jurysek:
> Could you please merge the master branch? There was an issue with protobuf messing with the install and unit tests. The config should pass the tests also, let's see..

I did, but it didn't trigger a new test run..

jurysek:
> > Could you please merge the master branch? There was an issue with protobuf messing with the install and unit tests. The config should pass the tests also, let's see..
> 
> I did, but it didn't trigger a new test run..

Ah, actually it did, but the test failed again on the same issue.

vuillaut:
The CI is bugged, but the config is valid.
I will merge and run the prod.

vuillaut (approved):

vuillaut:
@jurysek for the record, how did you generate the custom config file please?

vuillaut:
@jurysek sorry I overlooked the config during review and did not realise you limited it to DL1 processing
Could you explain why?

jurysek:
> @jurysek for the record, how did you generate the custom config file please?

I did manually, considering that I only need the first two stages to be run.

jurysek:
> @jurysek sorry I overlooked the config during review and did not realise you limited it to DL1 processing Could you explain why?

This PR is to get DL1 for the new testing diffuse gammas, which haven't been available until now. These I will further process using my RFs trained on tunned MC, which I produced before lstmcpipe was accustomed for allsky production. In PR #306, Giorgio is asking for training RFs in the lstmcpipe framework, so once this is done, one can also use those RFs to process the new testing diffuse gammas in order not to double the work.

vuillaut:
> These I will further process using my RFs trained on tunned MC, which I produced before lstmcpipe was accustomed for allsky production. 

This could very well be added to the present production, the fact that the RFs were not produced with lstmcpipe does not matter, we only need their path on the cluster.

jurysek:
> > These I will further process using my RFs trained on tunned MC, which I produced before lstmcpipe was accustomed for allsky production.
> 
> This could very well be added to the present production, the fact that the RFs were not produced with lstmcpipe does not matter, we only need their path on the cluster.

Yes, but I would like first to check if the tunning parameters are still the same, so I need to check DL1 first. For me the easiest way is just to get standard DL1 from lstmcpipe and then proceed with my own scripts.

jurysek:
> Prod is finished:
> 
> ```
> /fefs/aswg/data/mc/DL1/AllSky/20220527_src2_diffgamma/TestingDataset/GammaDiffuse/
> ```

Great, thanks!

PR #323, opened 2022-06-02
fix CI  validate configs

vuillaut:


vuillaut:
```
Error: Can't use 'tar -xzf' extract archive file: /home/runner/work/_actions/_temp_ca53292a-3b24-4a68-bd1d-aaceb4d3a773/5f825031-eb59-49ca-83e5-f5491f6732d2.tar.gz. return code: 2.
``` 
???

garciagenrique:
If prod5 configs they need to be generated by a script, we should document this somewhere. Otherwise looks good to me.

garciagenrique:
Although not clear why we leave some entries on the config file and why we take out others

vuillaut:
> If prod5 configs they need to be generated by a script, we should document this somewhere. Otherwise looks good to me.

The documentation already says:
_To generate your lstmcpipe configuration file, use lstmcpipe_generate_config command. If the type of production you want is not listed in the existing ones, you may create your own PathConfig class from an existing one,_

And the list of implemented classes, from the help:
```
['PathConfig', 'PathConfigProd5Trans80', 'PathConfigProd5Trans80DL1ab', 'PathConfigAllSkyBase',
                        'PathConfigAllSkyTraining', 'PathConfigAllSkyTesting', 'PathConfigAllSkyFull', 'PathConfigAllSkyTrainingDL1ab', 'PathConfigAllSkyTestingDL1ab',
                        'PathConfigAllSkyFullDL1ab']
```

vuillaut:
> Although not clear why we leave some entries on the config file and why we take out others

I left the mandatory ones 🤔

PR #324, opened 2022-06-03
refactor pointings

### LINKED ISSUE #325: Training pointings not actually intersected
Original comment by @jsitarek in https://github.com/cta-observatory/lst-sim-config/issues/1#issuecomment-1144088330_

made me realise pointings are not actually well intersected, and these pointings that are different from Protons and GammaDiffuse actually have been included in the prods.

vuillaut:
Simplifies the way pointings are loaded and intersected between Protons and GammaDiffuse using astropy tables.
The `join_type` option also allows to load all pointings if needed.

Fixes #325

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -71,7 +71,7 @@ jobs:
         run: |
           source $CONDA/etc/profile.d/conda.sh
           conda activate ci
-          export filelist=${{ steps.verify-changed-files.outputs.all_changed_files }}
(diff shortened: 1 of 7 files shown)


garciagenrique:
Add astropy as dependency, please (and fix the version :-) )

vuillaut:
> Add astropy as dependency, please (and fix the version :-) )

what do you mean? it is already in the dep requirements

garciagenrique:
> 

Sorry, right, it is already included :D

PR #339, opened 2022-09-06
making sure CI runs only on yaml files

vuillaut:
Fixing current issues with config validation during CI


garciagenrique:
Thanks @vuillaut !

PR #336, opened 2022-09-02
20220902 crab test nodes

vuillaut:
# New Prod config

## Short description of the config

Generated as:

```
lstmcpipe_generate_config PathConfigAllSkyTesting --prod_id 20220902_crab_test_nodes --kwargs dec=dec_2276
```

Then edited manually to replace the path model with `/fefs/aswg/data/models/AllSky/20220518_allsky_dec2276_tuned/dec_2276/`

lstchain config copied from `/fefs/aswg/data/models/AllSky/20220518_allsky_dec2276_tuned/dec_2276/`

## Why this config is needed 

Config to produce the new test nodes using Crab dec training.
See https://github.com/cta-observatory/lst-sim-config/issues/49#issuecomment-1235392024

PR #340, opened 2022-09-06
Update CI.yml

garciagenrique:
CI syntax was not correct: https://github.com/cta-observatory/lstmcpipe/actions/runs/3002594699

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -67,14 +67,16 @@ jobs:
             production_configs/*/*.yaml
 #            production_configs/*/*.json
 # TODO: solve issue in github actions extracting sublist of yaml files, see below
+
+#          export filelist=`grep -P '\.yaml$|\.yml$' <<< ${filelist[*]}`
+# the above line works on ubuntu but fails on github action, not following the given pattern
+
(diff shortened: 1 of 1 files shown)


vuillaut (changes requested):
why moving this line here?
I'd prefer it to stay where it was, or in  the future this comment won't make sens.
I think the problem was the indent?

garciagenrique:
> why moving this line here? I'd prefer it to stay where it was, or in the future this comment won't make sens. I think the problem was the indent?

Exactly, the indent. Now, GitHub does not have CI validator, the only way I found to check that the CI file is valid was by editing it using GitHub (not the github.dev !), that underlines in red the lines that are not correct.  

Feel free to push to this PR in the moment you find a valid syntax.

vuillaut (approved):

vuillaut:
Here is what I meant.
If ok with you, you can merge.

garciagenrique:
> Here is what I meant. If ok with you, you can merge.

The editor did not like this option either, but if it works, of course I'm ok

garciagenrique:
FYI, testing it on #336

PR #345, opened 2022-10-06
20221006 src3 test nodes

SeiyaNozaki:
# New Prod config

## Prod_ID
20221006_src3_test_nodes

produce Test DL1 datasets including newly produced test MCs (`theta_23.630_az_100.758` and `theta_23.630_az_259.265`)
This production config includes only DL1ab and DL1 merge process

### DIFF
--- a/production_configs/20221006_src3_test_nodes/README.md
+++ b/production_configs/20221006_src3_test_nodes/README.md
@@ -1,0 +1,13 @@
+# 20221006_src3_test_nodes
+
+DL1ab with tuned configuration
+(DL1ab + merge DL1)
+
+lstchain config taken from 2022-06-01 prod
+
+
+## command-line:
+```
+lstmcpipe_generate_config PathConfigAllSkyTestingDL1ab --prod_id 20221006_src3_test_nodes --kwargs source_prod_id=20220902_crab_test_nodes dec=3476  
+```
+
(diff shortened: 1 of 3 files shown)


vuillaut (approved):
Hi @SeiyaNozaki 
Looks good to me.
Note that you used a version of lstmcpipe that is not a release (master ?). It's fine, this will work and I can retrieve the exact commit you used but in general, you are encouraged to use tagged versions, it will be much easier to re-run the prod if needed in the future.

SeiyaNozaki:
Hi @vuillaut 
> you are encouraged to use tagged versions,

ah ok, thanks... I will do so next time.
I was not sure how to deactivate `run_checker` flag which raised an error (because there are more simtel files now than before), so I just modified the code directly just to create the production config without this flag in the end...

vuillaut:
> Hi @vuillaut
> 
> > you are encouraged to use tagged versions,
> 
> ah ok, thanks... I will do so next time. I was not sure how to deactivate `run_checker` flag which raised an error (because there are more simtel files now than before), so I just modified the code directly just to create the production config without this flag in the end...

Thank for reporting the issue!
I have prepared #347 to fix this.
I will take care of this prod now.

vuillaut:
Ha but you left out these new nodes in the lstmcpipe config file, so it (rightfully) complains that it can't find the corresponding DL1 files.
Fix #347 seems to work find, I have regenerated the config using the command line you provided in the README (thanks for including it) and launched the production. I will let you know how it goes 🚀

SeiyaNozaki:
ah you're right... Thank you @vuillaut !

PR #349, opened 2022-10-07
dependency for dl1_dl2 is either merge or train pipe

### LINKED ISSUE #348: dl1_to_dl2 seems not to wait for dl1_merge

vuillaut:
Fixes #348


garciagenrique (approved):
A bit overkill the if conditions, but looks nice :)

vuillaut:
> A bit overkill the if conditions, but looks nice :)

Why overkill?
Could it be simplified?

PR #350, opened 2022-10-08
Fix 20221006

vuillaut:
Follow-up to #345


vuillaut:
@SeiyaNozaki I fixed the faulty config, the prod is done.

Issue #191, opened 2022-03-06
Some modification required for source-dependent MC production


chaimain:
Following the discussion in #188 we will need to add `--source-dep` option for generating IRFs, and maybe some more things, as the discussion with @SeiyaNozaki progresses, regarding producing source dependent MC production.

LukasNickel:
Choosing pointlike gammas for RF right?
Anything else?
Lets discuss it in the call

SeiyaNozaki:
I think we need to modify for source-dependent analysis:
https://github.com/cta-observatory/lstmcpipe/blob/master/lstmcpipe/stages/mc_train.py#L52-L53
to use point gamma MC as RF training input
and
https://github.com/cta-observatory/lstmcpipe/blob/master/lstmcpipe/stages/mc_dl2_to_irfs.py#L284
to add `--source-dep` flag if soure-dep analysis is selected

LukasNickel:
Pointlike gammas for all random forests. Is that a good idea?

Is it enough to put source dependent in the config and use the normal for dl1/dl2 or is there something missing?

SeiyaNozaki:
> Pointlike gammas for all random forests. Is that a good idea?

yes, this is the way for source-dependent analysis. It would make some bias for proton/electrom MC, but we can obtain better performance for point gamma source.

> Is it enough to put source dependent in the config and use the normal for dl1/dl2 or is there something missing?

It should be enough only to activate source-dep flag in the config.

vuillaut:
I believe this can now be easily supported by the new PathConfig logic.
Is there still a need @SeiyaNozaki @chaimain ?

vuillaut:
Hi @SeiyaNozaki 
With the AllSky production, what gamma do you use for training for the source-dep analysis?

SeiyaNozaki:
For all sky MC, I'm using diffuse gamma as RF training for source-dependent analysis (though I used point gamma before all sky MC).

vuillaut:
> For all sky MC, I'm using diffuse gamma as RF training for source-dependent analysis (though I used point gamma before all sky MC).

ha ok great, so #375 is in line with your processing and it also means there is no changes required in lstmcpipe to process source-dependent !!
You can request a prod for your source-dep analysis, the only difference is in the lstchain config.

PR #375, opened 2022-11-29
New production config 20221128_src5_dec3476_tuned

alvmas:
# New Prod config

## Prod_ID
20221128_src5_dec3476_tuned

dec_3476 with NSB tuning to src5 and src-dependent configuration.

## Why this config is needed 
Reprocess the data at declination 34.76 with version of lstchain v0.9.9 and with new tunning for new source: src5.


vuillaut (approved):

vuillaut:
Production running 🚀

vuillaut:
Hi @alvmas 
I overlooked something for that production: the source-dependent analysis is trained on point-source gammas, which was not the case here.
See  discussion here: https://github.com/cta-observatory/lstmcpipe/issues/191

PR #381, opened 2022-12-12
Produce RF with release v0.9.6 to verify new DL1 compatibility

FrancaCassol:
# New Prod config
lstmcpipe_config_2022-11-02_PathConfigAllSkyFullDL1ab.yaml
## Prod_ID
20221212_v0.9.6_src4_high_NSB_LZA
See production [https://github.com/cta-observatory/lstmcpipe/pull/356](https://github.com/cta-observatory/lstmcpipe/pull/356)

## Why this config is needed 
we want to verify eventual incompatibilities between lstchain version v0.9.9 and v0.9.6 concerning application of RFs to DL1b


vuillaut:
Hi @FrancaCassol 

Sorry, there is actually a bit more work here. If we roll with the same prod_id and the paths in the current config, the previous production will be overwritten, which we don't want.

You will need to regenerate the config with a new prod_id. 
Please install lstmcpipe v0.10.0 to do so (to avoid issues with the moved MC prod).
``` 
pip install lstmcpipe==0.10.0
```

The script used to generate the previous config is in the readme of your PR, so it should be very straightforward.

Don't forget to rechange the lstchain env to 0.9.6 in the generated config ;-)

vuillaut (changes requested):
The config must be regenerated

PR #388, opened 2022-12-15
Add full production with standard NSB/PSF for AGN analysis

mireianievas:
# New Prod config

Self check-list:

- [ x] I have checked the environment in the lstmcpipe config and it is the one used to analyse DL>1 data
- [ x] I have provided the command (in README), or script (in additionnal .py file) used to produce the lstmcpipe config

## Prod_ID

20221215_v0.9.12_base_prod

Standard NSB standard PSF for processing standard AGN dark data. 

Declination tracks needed:

- dec_2276 
- dec_3476 
- dec_4822 
- dec_6676 
- dec_931
- dec_min_413 

While some sources may require tuned NSB, I want to check myself how an incorrect NSB spoils the reconstruction 

## Why this config is needed 

The last full production (DL1 + models) that is available is from ~ May. Since then, mostly incomplete productions and tuned ones for specific sources.

### DIFF
--- a/production_configs/20221215_v0.9.12_base_prod/lstchain_config.json
+++ b/production_configs/20221215_v0.9.12_base_prod/lstchain_config.json
@@ -1,0 +1,303 @@
+{
+    "source_config": {
+        "EventSource": {
+            "allowed_tels": [
+                1
+            ],
+            "max_events": null
+        },
+        "LSTEventSource": {
+            "default_trigger_type": "ucts",
+            "allowed_tels": [
+                1
+            ],
+            "min_flatfield_adc": 3000,
+            "min_flatfield_pixel_fraction": 0.8,
+            "calibrate_flatfields_and_pedestals": false,
+            "EventTimeCalculator": {
+                "dragon_reference_counter": null,
+                "dragon_reference_time": null
+            },
+            "PointingSource": {
+                "drive_report_path": null
+            },
+            "LSTR0Corrections": {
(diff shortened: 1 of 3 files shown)


vuillaut:
What version (or how did you install it) of lstmcpipe did you use to generate the config?
The config shows a dev v0.1+ version.
Did you need to modify lstmcpipe? If not please use a tagged version (0.10 is the last one) so we can reproduce these results.

mireianievas:
Ok, I used the main repository, synced today. Let me update the pull request config file using a tagged version.

mireianievas:
Done.

vuillaut:
Another thing.
I notice you left `sin_az_tel` in the config. Be aware that all real data produced with lstchain <= v0.9.7 can't be analyzed that way. Is that ok ?

mireianievas:
But I can reprocess those old data, right? My idea is to have an homogeneous processing of the entire datasets with a new lstchain.

vuillaut:
> But I can reprocess those old data, right? My idea is to have an homogeneous processing of the entire datasets with a new lstchain.

You can, it's usually done by lstosa though.

vuillaut:
Merging and launching prod

vuillaut:
🚀

mireianievas:
@vuillaut I just learnt that the azimuth -> sin(az) change was recently reversed, so now it is decided we will use azimuth again. Shall I make, for the sake of consistency, a new pull request with that parameter in the model generation?

PR #389, opened 2022-12-20
Add full production with standard NSB/PSF for AGN analysis (az instead of sin_az in features)

mireianievas:
# New Prod config

Self check-list:

- [ x] I have checked the environment in the lstmcpipe config and it is the one used to analyse DL>1 data
- [ x] I have provided the command (in README), or script (in additionnal .py file) used to produce the lstmcpipe config

## Prod_ID

20221215_v0.9.12_base_prod_az

Standard NSB standard PSF for processing standard AGN dark data. Changing sin_az -> az in training features w.r.t the previous production request after clarification during the last analysis call. 

Declination tracks needed:

- dec_2276 
- dec_3476 
- dec_4822 
- dec_6676 
- dec_931
- dec_min_413 

While some sources may require tuned NSB, I want to check myself how an incorrect NSB spoils the reconstruction 

## Why this config is needed 

The last full production (DL1 + models) that is available is from ~ May. Since then, mostly incomplete productions and tuned ones for specific sources.


moralejo:
> Hi @mireianievas Could you explain (or point to an explanation) of why `az_tel` should be used instead of `sin_az_tel` please? (for reference, and for me since I missed that analysis call)
> 
> Also I think `az_tel` would be a bit more explicit in the PROD_ID (that means regenerating the config)

I think that must be a mistake, in the checklist az_tel is mentioned - and I think the idea is to use this to process DL1 including files created with lstchain earlier than 0.9.7

vuillaut:
The provided lstchain config here still uses `sin_az_tel` in all features.

production_configs/20221215_v0.9.12_base_prod_az/lstchain_config.json
```diff
@@ -0,0 +1,303 @@
...
+        "width",
+        "length",
+        "x",
+        "y",
+        "wl",
+        "skewness",
+        "kurtosis",
+        "time_gradient",
+        "leakage_intensity_width_2",
+        "sin_az_tel",
```
vuillaut, line 168: That is precisely what you wanted to change, no?

vuillaut (changes requested):
The provided lstchain config here still uses sin_az_tel in all features.

mireianievas:
Sorry @vuillaut @moralejo , I completely lost the track of this pull request with the Christmas break, grant application and other deadlines. 

I believe it is correct now? 

It should read az everywhere, definitely sin_az. About the label of the production, since I accidentally got the sin_az production without extra labels, I believe I am forced to add some label (_az) to this one.

mireianievas:
> Hi @mireianievas Could you explain (or point to an explanation) of why `az_tel` should be used instead of `sin_az_tel` please? (for reference, and for me since I missed that analysis call)
> 
> Also I think `az_tel` would be a bit more explicit in the PROD_ID (that means regenerating the config)

The idea is to be able to process all 'standard' data in a homogeneous way, including that before 0.9.7. I previously requested a sin_az production as I thought that change was not going to be reversed, which it apparently is: @moralejo can you confirm? .

vuillaut (approved):

PR #393, opened 2023-01-27
renamed prod 20230127_v0.9.12_base_prod_az_tel and rerun config generation

vuillaut:
# New Prod config

Self check-list:

## Prod_ID

20230127_v0.9.12_base_prod_az_tel

Continutation of #389 

## Why this config is needed 

@mireianievas I have changed the prod_id and rerun the config generation to include the new MC nodes (if we run a complete new prod, better have them).


vuillaut:
🚀

PR #407, opened 2023-06-01
20230601 v0.9.13 dec 4822 dec 6166

aaguasca:
# New Prod config

Self check-list:

## Prod_ID
20230601_v0.9.13_dec_4822_dec_6166

Create a production with lstchain-v0.9.13 with all testing nodes for dec_4822 and dec_6166 with sin_az_tel. The stage r0_to_dl1 is the only stage used because this production will be used as the base prod. to latter tune the NSB.

## Why this config is needed 
Processing of source 7 with MC processed with lstchain-v0.9.13 for dec. bands dec_4822 and dec_6166 with all available testing nodes.


vuillaut (approved):

vuillaut:
🚀

PR #409, opened 2023-06-02
Bug: fix workflow_step label in log files when dl1ab step is used

aaguasca:
When the lstmcpipe is run using the dl1ab step, the workflow_step label `r0_to_dl1` appears instead of the `dl1ab` in the log_reduced*.yml and log_lstmcpipe_*.yml files.

This PR should fix this bug.


vuillaut (approved):
Looks good, thanks!
Could you add yourself (as new author) to the `codemeta.json` file as well in the same PR?

PR #412, opened 2023-08-29
Add new production src5 with lstchain v0.10

alvmas:
20230829_src5_dec3476_v0.10_tuned 

# New Prod config

Self check-list:


alvmas:
Hi, I corrected the config file but it seems to be failing the PR as well.

vuillaut:
> Hi, I corrected the config file but it seems to be failing the PR as well.

this was a fluke, thanks for fixing the config, tests are now passing, I will launch the prod

PR #413, opened 2023-08-30
Test Python versions from 3.7 to 3.11

vuillaut:


.github/workflows/CI.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         # lstchain dependency
-        python-version: [3.7]
+        python-version: [3.7, 3.8, 3.9, 3.10, 3.11]
```
morcuended, line 14: current lstchain requires >=3.9

vuillaut: Thanks Daniel!
where do you see this requirement in lstchain? `setup.py` says nothing about python version

morcuended: in `setup.cfg`

vuillaut: Thanks!
Since this lstmcpipe version is still compatible with lstchain >= 0.9.7 I will test them all for now and remove former versions when requiring latest version of lstchain.

PR #415, opened 2023-09-01
Add configs for new allsky production using lstchain 0.10.4

LukasNickel:
# New Prod config

Self check-list:

All grid nodes with default settings and lstchain 0.10.4.

## Why this config is needed 

There have been datamodel-changes in the transition from lstchain 0.9 to 0.10, which is why a new production is in order.
Also old `lstchain` configs dont necessarily work with the new versions (e.g. due to floating point `verbose` values).

This request includes all declination lines and no NSB-tuning in order to act as a new baseline.
For the same reason all processing steps are included.
That should make it possible to directly compare the performance to the older `20230127_v0.9.12_base_prod_az_tel` production.

For Crab or other high-NSB sources, a `dl1ab` reprocessing or another prodcution will thus be needed.

## Notes

I personally can not make use of all the grid lines, but I figured it would be useful to have them around as most simulations right now are 0.9 and earlier and this makes life a bit more difficult.
In fact I don't even need the later stages of the pipeline, but since they are way cheaper than the DL0->DL1 steps (which I do need), I included them as well. Then we directly have a comparison of 0.9.12 to 0.10.4.

### DIFF
--- a/production_configs/20230901_v0.10.4_allsky_base_prod/README.md
+++ b/production_configs/20230901_v0.10.4_allsky_base_prod/README.md
@@ -18,3 +18,13 @@ For the same reason all processing steps are included.
 That should make it possible to directly compare the performance to the older `20230127_v0.9.12_base_prod_az_tel` production.
 
 For Crab or other high-NSB sources, a `dl1ab` reprocessing or another podcution will thus be needed.
+
+## Command to produce
+```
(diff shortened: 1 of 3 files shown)


vuillaut:
Hi

Thank you very much for the request and detailed description.
Looks good to me, launching asap.

vuillaut (approved):

PR #433, opened 2023-10-18
remove 20230921 production folder

SeiyaNozaki:
This PR just removes 20230921 production folder, which had typo, just to avoid any confusion in the future. You can also remove corresponding MC files from the cluster. 

Thank you!


vuillaut (approved):
prod has been deleted

PR #442, opened 2024-01-31
20240131 v0.10.5 allsky crab base

vuillaut:
# New Prod config

Self check-list:

- [ ] I have checked the lstchain config, in particular for:
  - [ ] `az_tel` instead of `sin_az_tel` if data to be analyzed have been produced with lstchain <= v0.9.7
  - [ ] "increase_nsb" and "increase_psf" are provided in "image_modifier" (if used)

Base production for crab line with all stages

## Why this config is needed 

To analyse data with lstchain v0.10.5

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -14,3 +14,4 @@ dependencies:
   - lstchain
   - pyirf
   - ctaplot
+  - scipy~=1.11.4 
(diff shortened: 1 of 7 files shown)


vuillaut:
replaced with #443

PR #447, opened 2024-02-06
ask 20240126_v0.10.4_src6_dec2276_tuned production 

alvmas:
# New Prod config

Self check-list:

New Prod Config

## Prod_ID
20240126_v0.10.4_src6_dec2276_tuned

Config for the processing of MC at dec_2276 with NSB tuning to src6 and src-dependent configuration using a specific NSB tunning for the source.

## Why this config is needed 
Reprocess the data at declination 22.76 with version of lstchain v0.10.4 and with new tunning for new source: src6.


vuillaut:
Prod started 🚀  @alvmas 
The cluster is quite busy at the moment with the school and many analyses running so it could take a few days. I will let you  know.

PR #451, opened 2024-02-09
path config for a train-test analysis from existing prod

vuillaut:
New Path Config to restart from DL1b files and do train+test


vuillaut:
Tested with 
```
lstmcpipe_generate_config PathConfigAllTrainTestDL1b --dec_list dec_2276 dec_931 --prod_id GREATPROD --kwargs source_prod_id=20240131_allsky_v0.10.5_all_dec_base 
```

PR #463, opened 2024-04-30
add new production for src3 with tuned NSB for moon data analysis

joteros:
# New Prod config

Self check-list:

- [X] I have checked the lstchain config, in particular for:
  - [X] `az_tel` instead of `sin_az_tel` if data to be analyzed have been produced with lstchain <= v0.9.7
  - [X] "increase_nsb" and "increase_psf" are provided in "image_modifier" (if used)
- [X] I have checked the environment in the lstmcpipe config and it is the one used to analyse DL>1 data
- [X] I have provided the command (in README), or script (in additionnal .py file) used to produce the lstmcpipe config

## Prod_ID
20240426_v0.10.9_src3_dec3476_4822_tunedNSB

## Short description of the config
Config for a source with dec within 3476/4822 and NSB tuning:

    "image_modifier": {
       "increase_nsb": true,
       "extra_noise_in_dim_pixels": 6.974,
       "extra_bias_in_dim_pixels": 1.825,
       "transition_charge": 8,
       "extra_noise_in_bright_pixels": 4.195
    }

lstmcpipe_generate_config PathConfigAllSkyFull --prod_id 20240426_v0.10.9_src3_dec3476_4822_tunedNSB --dec_list dec_3476 dec_4822

## Why this config is needed 
See wiki page (src 3)
https://www.lst1.iac.es/wiki/index.php/MC_analysis_and_IRF_production#lstmcpipe_productions

PR #464, opened 2024-04-30
Update versions of actions

morcuended:
There seems to be a problem with codecov step, something related to the repository token `Please upload with the Codecov repository upload token to resolve issue` says in the logs.

I see that the version of codecov is outdated:it was using v1, (current latest version is v4). This may also help solving the problem


morcuended:
@vuillaut, this PR made codecov work. So maybe this will solve the error in #463

vuillaut (approved):

vuillaut:
Thank you @morcuended !!

PR #484, opened 2024-09-06
reworked config to start from DL1 and add slurm options to better man…

vuillaut:
Hi @abhishekabhi063 
I remade the config to use the DL1ab step, no need to restart from R0 here.
I also added a few options that will smooth things regarding cluster usage.


vuillaut:
The production has been scheduled ⏳ 
(but be aware the cluster is very much used so it could take ~1 week)

vuillaut:
Hi @abhishekabhi063 
The production is done under ID `20240906_v0.10.9_GC_dec_min_2924_tuned`.

abhishekabhi063:
Thank you

PR #515, opened 2025-02-12
update setup conda to v3 in CI

vuillaut:


morcuended:
maybe you can use `mamba-org/setup-micromamba@v2`instead of conda setup

PR #520, opened 2025-06-17
First version of the of the 20250616_v0.11.0_LZA_investigation proces…

mdebony:
# New Prod config

Self check-list:

- [x] I have checked the lstchain config, in particular for:
  - [x] `az_tel` instead of `sin_az_tel` if data to be analyzed have been produced with lstchain <= v0.9.7
  - [x] "increase_nsb" and "increase_psf" are provided in "image_modifier" (if used)
- [ ] I have checked the environment in the lstmcpipe config and it is the one used to analyse DL>1 data
- [ ] I have provided the command (in README), or script (in additionnal .py file) used to produce the lstmcpipe config

## Why this config is needed 

The objectives of this productions is to process test MC simulations with different atmosphere profile to assess the impact of it at LZA.

PR #534, opened 2025-12-01
20251126_v0.10.12_lza_investigation

mdebony:
# New Prod config

Self check-list:

## Why this config is needed 

This a follow-up of #520, seasonal MC are currently being produced for the Boomerang paper and the objectives of this prod is to check if corsika simulation with flat atmosphere are a good approximation at 60° of zenith angle.

Configuration copied from the fixed version of 20250616_v0.10.12_LZA_investigation


vuillaut:
Hi @mdebony 
Could you create the branch on this repo for the tests to be able to run please?

mdebony:
Ok I will create a new PR

PR #535, opened 2025-12-08
20251126 v0.10.12 lza investigation

mdebony:
# New Prod config

Self check-list:

## Why this config is needed 

This a follow-up of #520, seasonal MC are currently being produced for the Boomerang paper and the objectives of this prod is to check if corsika simulation with flat atmosphere are a good approximation at 60° of zenith angle.

Configuration copied from the fixed version of 20250616_v0.10.12_LZA_investigation


vuillaut:
Hi @mdebony 
There are a few things I don't understand here.
There is a conflict between the `.py` files and the config files you provided. Did the `.py` files help you produce the config files?
- there is a script to change the number of estimators in the `lstchain_config.json` file but it has not been applied.
- even if it was, you are not requesting to train a model here, so changing the number of estimators would not do much.

mdebony:
Hi @vuillaut ,
The configurations is copied and modified from 20250616_v0.10.12_LZA_investigation which was itself copied from the last standard LST productions. I guess it's a useless remain that was not spotted before (I've removed it).
It's intended to not have any training and to use the standard RF.

vuillaut (approved):

vuillaut:
Thank you for the explanation

